PR #21, opened 2019-05-11
Lib3h stub (+ beginning of Lib3hProtocol rewrite)

neonphog:
- Pulls "JsonProtocol" from core
- Removes all the json serde references
- "JsonProtocol" -> "Lib3hProtocol"
- `serde_json::Value` -> `Vec<u8>`
- specifies "Address" as a byte array (Vec<u8>)

all open for discussion

----
Damien:

Hijacked this PR a bit to start stubbing Lib3h.
Added `NetworkEngine` trait to `lib3h_protocol`. It defines the common interface for all types of network modules to be used by the Lib3hWorker.
Only expected implementation is the `RealEngine` stubbed out in `lib3h`.
This engine handles incoming Lib3hProtocol messages with a FIFO.
I'm guessing its easier to debug and that we might want to have some control over CPU usage of the NetworkEngine by for example limiting the number of protocol messages to process per 'tick'.

RealEngine is configured at construction with a `RealEngineConfig` struct.


ddd-mtl:
FYI, me and David were thinking of redoing those data structs as compositions the next time we would have to refactor them.

lib3h_protocol/Cargo.toml
```diff
@@ -6,3 +6,5 @@ edition = "2018"
 
 [dependencies]
 holochain_logging = { path = "../logging" }
+serde = "=1.0.89"
+serde_derive = "=1.0.89"
```
ddd-mtl: I don't actually see this crates being used.

neonphog: `#[derive(Serialize, Deserialize` are from `serde_derive`

lib3h_protocol/src/lib.rs
```diff
@@ -0,0 +1,284 @@
+#![feature(try_from)]
+
+//! This module provides the api definition for working with lib3h
+
+#[macro_use]
+extern crate serde_derive;
+
+/// Opaque Address Bytes
+pub type Address = Vec<u8>;
```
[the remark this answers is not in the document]
neonphog: Agree, this change isn't correct. Part of me wants to take the raw bytes of the underlying hash or public key... but then if we want to represent it in debugging or whatnot we'll have to re-do the reed-solomon in the case of HCIDs... so we should probably just treat the addresses as strings as we have been...

lib3h/Cargo.toml
```diff
@@ -8,3 +8,6 @@ edition = "2018"
 holochain_logging = { path = "../logging" }
 holochain_lib3h_protocol = { path = "../lib3h_protocol" }
 holochain_sodium = { path = "../sodium" }
+serde = "=1.0.89"
+serde_derive = "=1.0.89"
+serde_json = { version = "=1.0.39", features = ["preserve_order"] }
```
neonphog: I think it'd be awesome if we could remove any json assumptions from lib3h. lib3h itself doesn't need to serialize to json as it'll just be receiving the enums directly.

lib3h/src/p2p_protocol.rs
```diff
@@ -0,0 +1,10 @@
+/// Enum holding all message types in the 'network module <-> network module' protocol.
+/// TODO
+#[derive(Debug, Clone, PartialEq)]
+pub enum P2pProtocol {
```
neonphog: let's not include this yet, we're not ready to tackle p2p messaging

lib3h/src/network_engine.rs
```diff
@@ -0,0 +1,14 @@
...
+/// Common interface for all types of network modules to be used by the Lib3hWorker
+pub trait NetworkEngine {
+    /// Start network communications
+    fn run(&self) -> Lib3hResult<()>;
+    /// Stop network communications
+    fn stop(&self) -> Lib3hResult<()>;
+    /// Terminate module. Perform some cleanup.
+    fn terminate(&self) -> Lib3hResult<()>;
+    /// Handle some data message sent by the local Client
+    fn serve(&self, data: Lib3hProtocol) -> Lib3hResult<()>;
```
neonphog: I was imagining bubbling this all up one level. That is, I think the api to the networking module might be a little cleaner if we make the net crate in core deal with the worker / channel stuff. We could move this trait into lib3h_protocol and instead of `serve` have something like:

```rust
pub type DidWork = bool;

pub trait NetworkEngine {
    /// post a protocol message from core -> lib3h
    fn post(&mut self, msg: Lib3hProtocol) -> Lib3hResult<()>;

    /// a single iteration of the networking process loop
    /// (should be called frequently to not cpu starve networking)
    /// returns a vector of Lib3hProtocol events core will need to handle
    fn poll(&mut self) -> Lib3hResult<(DidWork, Vec<Lib3hProtocol>)>;
}
```

This is the pattern I arrived at for the WSS transport stuff: https://github.com/holochain/holochain-rust/blob/develop/net/src/ipc/transport/mod.rs

We certainly don't have to go this way if others have other ideas, but this is what I was talking about in terms of pushing the complexity of dealing with asynchronous stuff outside of lib3h.

lib3h/src/mock_engine.rs
```diff
@@ -0,0 +1,134 @@
+use std::sync::mpsc;
+
+use crate::{network_engine::NetworkEngine, p2p_protocol::P2pProtocol};
+use holochain_lib3h_protocol::{protocol::Lib3hProtocol, Lib3hResult};
+
+/// Lib3h's 'mock mode' as a NetworkEngine
+pub struct MockEngine {
```
neonphog: I *think* we can do this as part of the real lib3h networking engine.

Even for real networking, if there are multiple agentIds in the same DNA in the same conductor, and they want to send messages / make requests of each other, we'll want to short-circuit that communication path and redirect it back to core.

Some things will be a little hacky / stub-y like until we have a dht / gossip module, we'll also need to reflect DHT fetch requests back as well. But as we iterate and add new modules, we can slowly add real functionality to it.

Thoughts? I just think it'd be cool to only have one real module to maintain going forward.

lib3h/src/p2p_protocol.rs
ddd-mtl: ok

lib3h/src/network_engine.rs
ddd-mtl: I like that. I'd rename `poll` to `process` though since thats the main intent of calling this function and fits the engine mental model.

lib3h/Cargo.toml
ddd-mtl: I like banning serde (json). Felt like needed it for handling network config but doing it manually in net crate is not hard,

lib3h/src/mock_engine.rs
ddd-mtl: Fine by me.

ddd-mtl:
@neonphog I have taken in account your suggestions. Let me know if it works for you.

lib3h_protocol/src/data_types.rs
```diff
@@ -0,0 +1,195 @@
+use crate::Address;
+
+/// Tuple holding all the info required for identifying a metadata.
+/// (entry_address, attribute, content)
+/// TODO: Figure out if we keep this
+pub type MetaTuple = (Address, String, Vec<u8>);
```
dymayday: Just wondering why using a tuple here instead of a `named-field struct`, wouldn't it be easier to use later on and less prone to error if one day we decide to add some element to this data structure ? Furthermore, a `named-field struct` will actually carry out the purpose of its content within it so it will be easier to understand what this data structure do.
What do you think ?

lib3h_protocol/src/lib.rs
```diff
@@ -1 +1,11 @@
+//! This module provides the api definition for working with lib3h
 
+pub mod data_types;
+pub mod network_engine;
+pub mod protocol;
+
+/// Opaque Address Bytes
+pub type Address = Vec<u8>;
+
+use failure::Error;
```
dymayday: Can we use a custom `error` implementation here ? I noticed some time ago that `Failure` ,despite being super useful, suffer big performance issues, especially on a happy paths (tho I suppose from the use case later on the code right ?)

From the official Failure documentation we can read:

> The `Error` type allocates. There are cases where this would be too expensive. In those cases you should use a custom failure.

And [here is reddit thread about those perf issues] (https://www.reddit.com/r/rust/comments/7te8si/personal_experience_can_cause_hard_to_notice/)

I am happy to implement the custom error myself if it can help you :)

What do you guys think ?

lib3h/src/real_engine.rs
```diff
@@ -0,0 +1,146 @@
...
+    /// Identifier
+    name: String,
+}
+
+impl RealEngine {
+    /// Constructor
+    pub fn new(config: RealEngineConfig, name: &str) -> Lib3hResult<Self> {
+        Ok(RealEngine {
+            config,
+            inbox: VecDeque::new(),
```
dymayday: Do we have any idea of the size of this queue ? Maybe we could use [VecDeque::with_capacity(usize)](https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.with_capacity) in order to help the compiler reserving some space on the heap. It's not absolutely required, just throwing some thoughts around :)

lib3h_protocol/src/data_types.rs
ddd-mtl: No need for a struct for its current use, and this might get removed in the future so I would like to keep it as simple as possible.

lib3h/src/real_engine.rs
ddd-mtl: Don't know the size we want yet, or if we even want to hard limit it.
Once we do, I'm all for reserving the memory space upfront 👍

lib3h_protocol/src/lib.rs
neonphog: I needed to move SodiumError into lib3h_protocol to integrate with core: https://github.com/holochain/lib3h/pull/57 - let me know how that fits into you guys' thoughts

ddd-mtl: `SodiumError into lib3h_protocol` sounds like a bad idea to me. What's the reason for that?

neonphog: If you have other ideas, totally let me know!

HolochainError needs a `From` defined for SodiumError (now Lib3hError). Previously, this was done in the sodium crate, requiring a dependency of the core_types crate. I don't think we want any crates in lib3h to depend on crates in core, and we can't make core_types depend on sodium, because core types needs to compile to WASM... hence adding it to lib3h_protocol which can be a dependency of both core_types and sodium.

lib3h/src/real_engine.rs
dymayday: It does limit anything, it just reserves the minimum amount we think we might need. And if it happens to be too small, the allocated size on the head will grow according to our need. But I see your point don't worry, and I think it's better to let it this way, so the compiler can optimize what it can better than us at the moment.

lib3h_protocol/src/lib.rs
ddd-mtl: I think this needs F2F discussion. Aren't we going to create a new crate for common types?
I still believe Core should break up `core_types` into WASM and non-WASM crates.

dymayday: Following the discussion on Mattermost, I'm writing here my proposal for a replacement of `Failure` for a custom `error` handling.

I would like for performance reason using a custom error handling like the one implement here, in [rust-csv](https://github.com/BurntSushi/rust-csv/blob/master/src/error.rs).

It requires a little bit of work but it will let us have a nice and fine grained error handling.

neonphog:
Approve @ddd-mtl - :+1: I can't do the github one since I opened this PR

PR #22, opened 2019-05-11
add makefile with rust nightly override matching core

neonphog:
- just a helper to get those who don't want to run core's nix-shell up and running


neonphog:
@struktured 

> I think this should work for those inside a nix-shell or outside one, ideally. The only reason it failed for me at the moment in a nix shell is the `rustup` step. Thus, I suggest one of the following options:

Good point. I was thinking of this as an either or... but we should totally make it work within the nix-shell as well. Alas, rustup (as a dependency manager) is incompatible with the philosophy of nix, but I think I can have the Makefile check the Rust toolchain version, and if it doesn't match check if rustup exists to set it, and if not, just bail.

neonphog:
@struktured - Makefile updated per comments, let me know what you think. I can't test it fully... even running `nix-shell --pure` the rustup from my system still seems to be available...

struktured:
Builds fine for me under nix-shell without any hacks.

neonphog:
@zippy - yep, I think that's the idea, either you run it from holochain-rust's nix-shell or you have rustup installed on your system (you can also install rustc, cargo, and rustfmt manually without rustup if you want)

@here - I'm taking the RFC off this, and merging, please feel free to bring up any concerns, we can address them in follow-ons

thedavidmeister:
all you need to get "core's nix shell up and running" is to run this:

`nix-shell https://holochain.love`

PR #23, opened 2019-05-11
move sodium crate over from core

neonphog:
- No changes, except removing a `HolochainError` from impl for sodium error.

I have a slight worry about keeping this implementation in sync with the one in core until we are set up to publish to crates.io and remove the core version... but we also haven't been changing sodium too much lately.


neonphog:
@struktured 

> What are the implications of changing HolochainError to SodiumError? Do we have to modify holochain core to make it compatible with this api change?

https://github.com/holochain/holochain-rust/blob/develop/sodium/src/error.rs#L17-L25
^^ was removed.

I *think* that `From` can be done in the `HolochainError` module too, so when we do the crates.io dependency swap, we'll have to add that.

sodium/README.md
```diff
@@ -0,0 +1,57 @@
...
+```
+
+## Contribute
+
+Holochain is an open source project.  We welcome all sorts of participation and are actively working on increasing surface area to accept it.  Please see our [contributing guidelines](https://github.com/holochain/org/blob/master/CONTRIBUTING.md) for our general practices and protocols on participating in the community.
+
+## License
+[![License: Apache-2.0](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](https://www.apache.org/licenses/LICENSE-2.0)
+
+Copyright (C) 2019, Holochain Trust
```
zippy:
```suggestion
Copyright (C) 2019, Holochain Foundation
```

I need to change this all over the place, but lets start here.

PR #24, opened 2019-05-11
copy websocket code over from core

neonphog:
Open Questions:
- [ ] - this was labeled `Transport` in core... but I've been calling it `Connection` in my interface documents... anyone have a preference or another suggestion?
- [ ] - How do we want to handle traits, directory structure and naming convention-wise? Here the trait is in the module `Transport` and the specific implementation is a sibling `TransportWss`... but we could also have a `traits` or `interfaces` directory, and name it `TransportTrait` or `TransportInterface`.


struktured:
> - [ ] - this was labeled `Transport` in core... but I've been calling it `Connection` in my interface documents... anyone have a preference or another suggestion?

The [transport layer](https://en.wikipedia.org/wiki/Transport_layer) is typically more general than the concept of a connection.  So if we intend to extend out the transport module with reliability and connection pooling etc. it may be best to call the module `transport`, but some other module (or crate) can be called `connection` that the transport module needs to establish connections.

lib3h/src/transport/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+/// events that can be generated during a connection `poll()`
+#[derive(Debug, PartialEq, Clone)]
+pub enum TransportEvent {
+    TransportError(TransportId, TransportError),
+    Connect(TransportId),
+    Message(TransportId, Vec<u8>),
+    Close(TransportId),
+}
+
+/// represents a pool of connections to remote nodes
```
struktured: This right here implies calling transport a connection instead would be a bit misleading.

struktured:
> [ ] - How do we want to handle traits, directory structure and naming convention-wise? Here the trait is in the module Transport and the specific implementation is a sibling TransportWss... but we could also have a traits or interfaces directory, and name it TransportTrait or TransportInterface.

This probably requires a discussion in standup/overflow to reach a consensus. Personally, I really dislike encoding type information in the names of entities. Regarding the directory structure, that is a more nuanced discussion as it depends on how much crate level factorization we want between traits/interfaces and implementations. In the short term I would maintain the existing patterns implied by the net crate in core until we can talk about it in a more detailed capacity.

ddd-mtl:
100% agreement with @struktured. Especially the disliking of `encoding type information in the names of entities`.

Issue #29, opened 2019-05-14
[REPO] add logging with configurable tagging and color that integrates with hc-core logging


jmday:
As a lib3h client, I can use a logger with configurable tagging and color so that I can efficiently debug network activity (and integrate my own logging with network logging)

Another wording:

As a developer, I wand to see networking related logging so that I can receive feedback on the state of the system.

- [x] - Simple macros so adding logging to modules is trivial
- [ ] - Configurable output channel (allow stdout / stderr / file)
- [x] - Configurable output filter (supports filter / coloring functionality currently in use by holochain-rust)

- [x] SPIKE: Given net crate’s tweetlog && core context’s logger, what is something that solves requirements && hopefully is fairly idiomatic to rust
- [x] RFC comprehensive logger solution to both core and networking teams
- [ ] Implement RFC logger solution

dymayday:
This PR #34  addresses this issue

ddd-mtl:
Maybe use a HackMD to spec this out?
The things I would like:
 - Multiple output channels at a time with different filters.
 - Seperating logLevels from tagging because logLevel should be something understandable by the compiler so it can strip out trace and debug logs when in production.
 - Possibility to add custom suffixs with runtime data to log outputs. (i.e. Adding PID to log output so we can discern instances in multinode scenarios).

zippy:
I've started: https://hackmd.io/MP5F3UhSTp2iPk37Cwa-fw

PR #34, opened 2019-05-15
Logging impl

dymayday:
This is the first implementation of the **asynchronous** `Logger`.

Related to this issue #29 

From the top of my head, here is a list of the things this logger will implement:
- [x] Simple logging using macros
- [x] Use config file
- [x] Log to file


logging/examples/features.rs
```diff
@@ -0,0 +1,12 @@
...
+pub fn main() {
+    let _guard = Logger::new();
+
+    trace!("logging a trace message");
+    debug!("debug values"; "x" => 1, "y" => -1);
+    info!("some interesting info"; "where" => "right here");
+    warn!("be cautious!"; "why" => "you never know...");
+    error!("wrong {}", "foobar"; "type" => "unknown");
+    crit!("abandoning test");
+}
```
struktured: @dymayday I find the the call instantiating the logger here a bit suspect. I feel either the method `new()` is misleading and instead should be named to imply a singleton pattern or that we register a logger reference with the macro system in a more explicit manner. Your thoughts?

logging/src/lib.rs
```diff
@@ -0,0 +1,81 @@
...
+/// This struct is used as a wrapper around the `slog` crate for our customized logger.
+/// This has to be instantiate at the top of the user program in order to keep
+/// the logging thread alive.
+pub struct Logger(slog_scope::GlobalLoggerGuard);
+
+impl Logger {
+    pub fn new() -> Self {
+        Logger::term_logger()
+    }
+
```
struktured: If this is a singleton it probably shouldn't be named `new()`. Is it setup / teardown style? Like an `init()` method? Or more like a wrapper function around a singleton? Eg. `get()` or `instance()` or `try...()`?

dymayday: I wanted to stick to idiomatic Rust with a `new` constructor, but indeed, calling a `init()` is probably more accurate, thanks for your input on this.

logging/src/lib.rs
```diff
@@ -1 +1,351 @@
...
+        let tmp_file = NamedTempFile::new().expect("Fail to create temporary file.");
+        let tmp_file_path = tmp_file.into_temp_path();
+        let tmp_file_path = tmp_file_path
+            .to_str()
+            .expect("Fail to convert 'PathBuf' to 'String'.");
+
+        let _guard = Lib3hLogger::init_file_logger(&tmp_file_path);
+        info!("{}", dumped_log_string);
+
+        // // This is the messed up part...
```
zippy: These comments should be fixed or removed.

neonphog:
Closing this, feel free to re-open if needed.

PR #38, opened 2019-05-16
Circleci

neonphog:
@thedavidmeister @duffybelfield

I'm sure there are better ways of doing a couple of these steps, but this is a working POC: https://circleci.com/gh/holochain/lib3h/9

- [x] - ~~resource_class: large?? I just copied this from the holochain-rust config, but seems like we could probably get away with smaller for the time being?~~ - switched to small

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,5 +11,5 @@ jobs:
   steps:
    - checkout
    - run: nix-shell -p git --run 'git clone https://github.com/holochain/holochain-rust.git'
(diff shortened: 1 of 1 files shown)


neonphog:
@duffybelfield says:
> looks good, let's get it working first then we can worry about making it prettier!

I'm gonna go ahead and merge. This is something we'll need to keep iterating on.

thedavidmeister:
@neonphog looks good to me, this is motivating me to nixify the fmt install properly using the approach from wasm bindgen :)

PR #56, opened 2019-05-20
clippy according to Montreal accord

### LINKED ISSUE #39: add clippy to Makefile / CI per Montreal accord

neonphog:
RESOLVES #39

```
cargo clippy -- \
  -A clippy::nursery -A clippy::style -A clippy::cargo \
  -A clippy::pedantic -A clippy::restriction \
  -D clippy::complexity -D clippy::perf -D clippy::correctness
```

clippy is no longer supported via `cargo install` and `rustup` is not supported in the nix environment... soooo I moved away from nix-shell for CI for now. I know @thedavidmeister was looking into bundling tools within the nix shell, so things like clippy would be available. Once that work is complete we can look at using nix again in CI. (unless someone has an idea how to fix this in the short term)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,11 +5,10 @@ jobs:
  # is actually unit tests
  build:
   docker:
-   - image: holochain/holochain-rust:latest
+    - image: rust:1.34.2-slim-stretch
   resource_class: small
   working_directory: ~/repo
   steps:
-   - checkout
-   - run: nix-shell -p git --run 'git clone https://github.com/holochain/holochain-rust.git'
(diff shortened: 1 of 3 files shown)


neonphog:
@ddd-mtl 

> Should we also add a rustfmt.toml file like in hc-rust so that Rust updates don't mess around with our code when rustfmt defaults changes?

I like that idea, I'll do that as a separate PR tomorrow morning

thedavidmeister:
@neonphog kk, lemme look into renixifying clippy (probably just needs to be built from the repo) then we can add it that way

PR #57, opened 2019-05-20
Introduce `lib3h_crypto_api` with error data types as an initial stub for crypto abstraction

neonphog:
@ddd-mtl @dymayday - Tagging you guys because of the error handling discussion in https://github.com/holochain/lib3h/pull/21 - let me know if this structure fits into what you're thinking.

These changes are needed for the holochain-rust integration: https://github.com/holochain/holochain-rust/pull/1416


lib3h_protocol/src/error/mod.rs
```diff
@@ -0,0 +1,41 @@
+//! Lib3h API Error Enum
+
+pub mod sodium_error;
+
+use sodium_error::SodiumError;
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub enum Lib3hError {
```
dymayday: So, from what I can read, `lib3h_protocol` is responsible for the `error` handling of the whole crate ? Not the root `lib3h` ?

neonphog: Discussed IRL. I'm going to move this into a new `holochain_crypto_api` crate :+1:

PR #61, opened 2019-05-21
Initial draft of dht & p2p modules.

ddd-mtl:
Copied over DHT and P2p interface from n3h as traits and followed the same "post & process" pattern for its first implementations.
Using `get_data()` for local queries, and `fetch_data()` for over the network queries.


neonphog:
@ddd-mtl - I haven't looked into this in detail yet, but I did notice the Transport trait - would it help to merge https://github.com/holochain/lib3h/pull/24 ? <- I got it updated with clippy / etc, just wasn't sure if everyone was cool with it as-is

I'll take a closer look at this PR tomorrow morning

ddd-mtl:
@neonphog Yeah, having that PR in would help. Or we could stub here a different interface more in line with the Connection interface we have in n3h.

neonphog:
@ddd-mtl - ok, the wss PR is merged. It put the `Transport` trait in lib3h, we'll probably want to move that into lib3h_protocol.

> Or we could stub here a different interface more in line with the Connection interface we have in n3h

Totally up for any suggested changes / updates to the interface :+1:

lib3h/src/p2p/p2p_trait.rs
```diff
@@ -0,0 +1,21 @@
...
+use holochain_lib3h_protocol::{DidWork, Lib3hResult};
+
+/// Composition of a Connection, Dht, RuntimeStore, PeerStore, and DataStore
+/// to provide a higher-level Connection interface.
+pub trait P2p {
+    /// Lifecycle
+    fn connect(&mut self, url: String) -> Lib3hResult<()>;
+    fn close(&self, peer_address: String) -> Lib3hResult<()>;
+    /// Comms
+    fn publish_reliable(&self, peer_list: Vec<String>, data: Vec<u8>) -> Lib3hResult<()>;
```
neonphog: Function parameters should accept references, and probably need to be `&mut self`:

```rust
fn publish_reliable(&mut self, peer_list: &[&Address], data: &[u8]) -> Lib3hResult<()>;
```

neonphog: Actually... if `Address` maps to `String` we probably need an `AddressRef` that maps to `str`:

```rust
pub type AddressRef = str;
fn publish_reliable(&mut self, peer_list: &[&AddressRef], data: &[u8]) -> Lib3hResult<()>;
```

lib3h/src/p2p/p2p_trait.rs
```diff
@@ -0,0 +1,21 @@
+use crate::p2p::p2p_event::P2pEvent;
+use holochain_lib3h_protocol::{DidWork, Lib3hResult};
+
+/// Composition of a Connection, Dht, RuntimeStore, PeerStore, and DataStore
+/// to provide a higher-level Connection interface.
+pub trait P2p {
```
neonphog: If we upgrade `Transport` to look more like [n3h's Connection](https://github.com/holochain/n3h/blob/master/lib/interface/connection-iface.js), then a "P2p" instance really is just an instance that implements both the `Transport` trait and the `Dht` trait perhaps?

lib3h/src/p2p/p2p_trait.rs
```diff
@@ -0,0 +1,21 @@
...
+    fn connect(&mut self, url: String) -> Lib3hResult<()>;
+    fn close(&self, peer_address: String) -> Lib3hResult<()>;
+    /// Comms
+    fn publish_reliable(&self, peer_list: Vec<String>, data: Vec<u8>) -> Lib3hResult<()>;
+    fn publish_unreliable(&self, peer_list: Vec<String>, data: Vec<u8>) -> Lib3hResult<()>;
+    fn request_reliable(&self, peer_list: Vec<String>, data: Vec<u8>) -> Lib3hResult<()>;
+    fn respond_reliable(&self, msg_id: String, data: Vec<u8>) -> Lib3hResult<()>;
+    /// Processing
+    fn post(&mut self, evt: P2pEvent) -> Lib3hResult<()>;
+    fn process(&mut self) -> Lib3hResult<(DidWork, Vec<P2pEvent>)>;
```
neonphog: Should we extract a trait, something like:

```rust
pub trait SteppingProcess<E> {
  fn post(&mut self, event: E) -> Lib3hResult<()>;
  fn process(&mut self) -> Lib3hResult<(DidWork, Vec<E>)>;
}
```

lib3h/src/p2p/p2p_trait.rs
```diff
@@ -0,0 +1,21 @@
...
+    /// Getters
+    fn id(&self) -> String;
+    fn advertise(&self) -> String;
```
neonphog: Mmm... these might be unique to P2p... so https://github.com/holochain/lib3h/pull/61/files#r286562038 isn't quite right, it's more like:

```rust
pub trait P2p: Transport + Dht + SteppingProcess<P2pEvent> {
  fn id(&self) -> String;
  fn advertise(&self) -> String;
}
```

... although P2pEvent then probably needs to be an enum that composes TransportEvent, DhtEvent, and any unique P2pEvent's?

lib3h/src/p2p/p2p_hackmode.rs
```diff
@@ -0,0 +1,218 @@
...
+use crate::{
+    dht::{dht_event::DhtEvent, dht_trait::Dht, rrdht::rrdht},
+    p2p::{p2p_event::P2pEvent, p2p_trait::P2p},
+    transport::{Transport, TransportEvent},
+    transport_wss::TransportWss,
+};
+use holochain_lib3h_protocol::{DidWork, Lib3hResult};
+use std::collections::VecDeque;
+
+pub struct P2pHackmode {
```
neonphog: Ack! I vote for not doing modes again. I think it should be possible to just code toward our ultimate end goal, stubbing out or hacking things where we need to, but with an eye toward coming back and implementing the correct version when we get to it.

lib3h/src/dht/dht_event.rs
```diff
@@ -0,0 +1,74 @@
...
+pub struct GossipToData {
+    pub peer_address_list: Vec<String>,
+    pub bundle: Vec<u8>,
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct PeerHoldRequestData {
+    pub peer_address: String,
+    pub transport: String,
+    pub data: Vec<u8>,
```
neonphog: Let's drop `data`... we don't have a specific use for it yet, and it may make more sense to add strongly-typed attributes rather than binary data.

lib3h/src/dht/dht_event.rs
```diff
@@ -0,0 +1,74 @@
...
+    pub timestamp: String,
```
neonphog: `u64` ms since epoch would be faster to compare than `String`

lib3h/src/dht/dht_event.rs
```diff
@@ -0,0 +1,74 @@
...
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub struct DataHoldRequestData {
+    pub data_list: Vec<Vec<u8>>,
```
neonphog: This is partly tied up in the discussion we're going to have tomorrow, but I do think we are going to need to store / sync some additional info outside the opaque content... so something like:

```rust
pub enum ContentType {
  Entry,
  Header,
  Meta,
}

pub struct DataHoldContent {
  pub type:ContentType,
  pub flags: Vec<(String, String)>, // arbitrary nv storage?? like: [("attribute", "link"), ("tag", "yay")]
  pub publish_ts: u64,
  pub content: Vec<u8>,
}

pub struct DataHoldRequestData {
  pub data_list: Vec<DataHoldContent>,
  pub data_address: Address,
}
```

lib3h/src/dht/rrdht.rs
```diff
@@ -0,0 +1,64 @@
+use holochain_lib3h_protocol::{Address, DidWork, Lib3hResult};
+
+use crate::dht::{
+    dht_event::{DhtEvent, PeerHoldRequestData},
+    dht_trait::Dht,
+};
+
+/// RoundAndRound DHT implementation
+#[allow(non_camel_case_types)]
+pub struct rrdht {}
```
neonphog: `RRDht`?

lib3h/src/p2p/p2p_trait.rs
ddd-mtl: We could. I thought about it, but I don't see what benefits that gives.

lib3h/src/p2p/p2p_trait.rs
ddd-mtl: ok

lib3h/src/p2p/p2p_hackmode.rs
ddd-mtl: hahaha. No worries, I just carried the name over from n3h. Gonna rename to `P2p` since that trait is actually useless.

lib3h/src/dht/dht_event.rs
ddd-mtl: ok. I was thinking of changing it to `hint: String`, but removing it seems even better.

lib3h/src/dht/rrdht.rs
ddd-mtl: `RrDht`

lib3h/src/dht/dht_event.rs
dymayday: I agree, type check is our best friend :)

ddd-mtl:
@neonphog Can we merge this in? Its still a draft, but I moved a lot of things around and I don't want to get merge conflicts...

lib3h/src/p2p/p2p_gateway.rs
```diff
@@ -0,0 +1,258 @@
...
+    transport_wss::TransportWss,
+};
+use lib3h_protocol::{Address, DidWork, Lib3hResult};
+
+/// Gateway to a P2P network.
+/// Enables Connections to many other nodes.
+/// Tracks distributed data for that P2P network.
+pub struct P2pGateway {
+    transport: Box<Transport>,
+    dht: Box<Dht>,
```
neonphog: I vote for using Generics rather than Trait Objects for composition, so we don't have to deal with all the restrictions and limitations of Trait Objects.

lib3h/src/dht/dht_trait.rs
```diff
@@ -0,0 +1,19 @@
...
+pub trait Dht {
+    /// Peer info
+    fn get_peer(&self, peer_address: String) -> Option<PeerHoldRequestData>;
+    fn fetch_peer(&self, peer_address: String) -> Option<PeerHoldRequestData>;
+    fn drop_peer(&self, peer_address: String) -> Lib3hResult<()>;
+    /// Data
+    fn get_data(&self, data_address: Address) -> Lib3hResult<Vec<u8>>;
+    fn fetch_data(&self, data_address: Address) -> Lib3hResult<Vec<u8>>;
+    /// Processing
+    fn post(&mut self, evt: DhtEvent) -> Lib3hResult<()>;
```
neonphog: I think it makes sense to me that `post` takes ownership of the DhtEvent.

Above, though, I think the `peer_address` params should be `&str` and the `data_address` params should be `&AddressRef`

lib3h/src/p2p/p2p_gateway.rs
ddd-mtl: You mean like this?
```Rust
pub struct P2pGateway<T: Transport, D: Dht> {
    transport: T,
    dht: D,
}
```

lib3h/src/dht/dht_trait.rs
ddd-mtl: oops forgot about that.

Issue #35, opened 2019-05-15
[REPO] SPIKE: Explore where to place crates with shared-use between HCCore and Networking.


jmday:
- [x] CAS
- [x] EAV (metadata storage vs indexing)
- [x] CAS / EAV implementations
- [x] libsodium
- [x] Logging
- [x] Lib3hProtocol

struktured:
> [ ] CAS

The CAS traits seem portable and abstract enough, as defined in [holochain-rust/core_types/src/cas/*](https://github.com/holochain/holochain-rust/tree/develop/core_types/src/cas). 

Holochain then has a `cas_implementations` crate which provides a file, in memory, and pickle db version. It would be straightforward to do *one* of the following:

 - move the traits _and_ implementations into a `XX` crate, shared by `holochain-rust` and `lib3h`
 - move just the traits into `XX` crate, leave `cas_implementations` in core
   - then we pick one of the databases discussed [here](https://hackmd.io/u3MgMRQDStCY5ra-1UQuFQ?both) as an implementation. Probably one of `rust-kv`, `sled`, or `sqlite3`.

The `XX` crate can either be
 - a new, separate `persistence` crate available from crates.io (w/separate repo?)
 - embedded in existing `lib3h` repo as a module
 - embedded in existing `lib3h` repo  and a _copy/fork_ of the code in `holochain-rust`, with decision of convergence or divergence from it deferred until a more strategic date

struktured:
> [ ] EAV (metadata storage vs indexing)

The EAV, as written currently, is not abstract enough for networking because the attribute is a enum very specific to subconscious metadata of holochain, such as links and crud. Also, the `Value` is typed as json string which may not be our best representation for networking. 

The concept of an EAV however seems an appealing one and without some sort of meta layer on top of the CAS it is not terribly useful. Perhaps we implement an EAV trait more suitable to networking but possibly generalize/merge it in the future with holochain-core (for instance, we could just make the attribute and value generic type parameters with extensible serde requirements). 

Will provide more thoughts and comments on this later- open to any suggestions in the meantime.

Issue #17, opened 2019-05-10
[IN-MEM] SPIKE - evaluate persistence options (like sqlite3) with an eye towards indexing / lru caching requirements for lib3h


neonphog:
We need to choose a database to persist content addressable data and have indexes over the data. This research should be time boxed and whatever product is chosen should be encapsulated within a persistence interface so we can try alternative implementations later.

The schema should consider different layers which have different key value associations.
 
- peer layer
    - key: machine id
    - value:```{transport_address:wss://...}```
- app/dna layer
    - key: tuple of dna hash and agent id
    - value: machine id

struktured:
[Hackmd](https://hackmd.io/u3MgMRQDStCY5ra-1UQuFQ?both) to build some notes on database choices.

PR #72, opened 2019-05-28
Persistence with string attributes

struktured:
This PR migrates over the `cas` and `eavi` modules from `holochain-core`. This consequently requires bringing over the `core_types_derive` crate to support the existing json serde framework. Thus `lib3h_persistence`, `lib3h_persistence_derive` crates are introduced. It makes no changes to the code except that 

 - `holochain_` occurrences are replaced by `lib3h_`
 - the `Attribute` enum is replaced by a type alias to `String`
 - validation of the attribute format is simplified and pushed downstream (previous code would validate link formats and such.)
 
This work was inspired and deduced by the SPIKE of issue #17. It addresses the persistence related crate  sharing between `holochain-core` and `lib3h` as noted in issue #35. The next step, assuming this PR is sound enough, is to migrate `cas_implementations` from `holochain-core` and make `holochain-core` build against it. 

Notes:

- `String` was favored over a generic trait to avoid complexity of lifetime parameters with `serde::Deserialize<'a>`. See https://circleci.com/gh/holochain/lib3h/95 and branch https://github.com/holochain/lib3h/tree/persistence-derive for details on this problem.
 - `String` was favored over an integer value to make it easier to support _dynamic_ attributes as needed by `holochain-core`. Eg. core has an attribute enum value called `LinkTag(String)` which would be difficult or error prone to map into an ordinal space.


ddd-mtl:
Agree with @neonphog. I don't know whats the smoothest path onward, but it seems to me we would want in the end some kind of `foundation`  or `basic_types` crate, which has all the basic, dependencyless code: `JsonString`, `HashString`, `Timeout` etc.
And the Holochain specific basic types would stay in the `holochain-rust` repo somehow.

struktured:
@neonphog @ddd-mtl I agree some of the holochain specific stuff should be moved out of this crate. I can do that before merging this in. As for the idea of shared / basic / foundational types, I would prefer to do that in another PR, if that's ok with the team- reason being it is probably a bit more involved to factorize them out.

struktured:
@neonphog @ddd-mtl I removed a bunch of holochain specific modules and changed `HolochainError` to `PersistenceError`. I then removed the variants from the error enum that weren't relevant to persistence or at least not used by persistence crate. Take a look and let me know if this is sufficient!

struktured:
Closing without merging in favor of #89  instead.

PR #89, opened 2019-06-02
Persistence api and implementation crates, with generic attributes 

struktured:
# Overview

This PR incorporates all of #72 but with the additional features:
 - attributes are no longer strings but instead are defined by [traits](https://github.com/holochain/lib3h/blob/persistence/crates/persistence/src/eav/eavi.rs#L31) and [trait bounds](https://github.com/holochain/lib3h/blob/persistence/crates/persistence/src/eav/eavi.rs#L113). 
 - the existing cas and eav implementations from `holochain_core/cas_implementations` has been factored out into a `lib3h_persistence_impl` crate (with file path `crates/persistence_impl`).

# Considerations
- > what is appropriate naming convention for the api crates? We have a `crypto_api` so should I rename the new `persistence` crate instead to `persistence_api`? 

 We decided on `persistence_api`.
 - > what is appropriate naming convention for impl crates? Eg. `holochain_core` has `cas_implementations` which is quite lengthy and not entirely accurate since it also has `eavi` implementations as well. I went with `persistence_impl` to both shorten it and make it general enough. Open to suggestions here.

Went with `persistence_mem`, `persistence_file`, and `persistence_pickle`.
 - some work is left to make this `crates.io` publishable (then holochain-core can start using it)
 
# Improvements
 - an [`Attribute`](https://github.com/holochain/lib3h/blob/persistence/crates/persistence/src/eav/eavi.rs#L31) trait has been introduced which is supposed to contain a base set of trait bounds all attributes must adhere to. It currently has quite a few and many of them shouldn't be required- instead these constraints can be pushed down into the storage implementation level.
 - It may even be possible to not require any `Attribute` trait at all- instead trait bounds are introduced as  compositions of existing and required traits only. The trade off is readability goes down- what we really want is a rust trait composition alias but they haven't mainlined that feature yet.


crates/persistence/src/error/mod.rs
```diff
@@ -0,0 +1,198 @@
...
+use std::{
+    error::Error,
+    fmt,
+    io::{self, Error as IoError},
+    option::NoneError,
+};
+
+//--------------------------------------------------------------------------------------------------
+// CoreError
+//--------------------------------------------------------------------------------------------------
```
ddd-mtl: dead comment

crates/persistence_impl/README.md
```diff
@@ -0,0 +1,8 @@
+# lib3h_persistence_impl
+
+Persistence implementations for lib3h and holochain. Provides content addressable storage (CAS) and entity attribute value (index) associative stores in three flavors.
+
+ - a pure, thread safe in memory storage
+ - file based storage using hiearchical directories to associate and query data
+ - [pickledb](https://github.com/seladb/pickledb-rs) for database backed storage
+
```
ddd-mtl: Same here: use our README "template"

crates/persistence_impl/src/eav/pickle.rs
```diff
@@ -0,0 +1,182 @@
...
+    collections::BTreeSet,
+    fmt::{Debug, Error, Formatter},
+    marker::{PhantomData, Send, Sync},
+    path::Path,
+    sync::{Arc, RwLock},
+    time::Duration,
+};
+use uuid::Uuid;
+
+const PERSISTENCE_INTERVAL: Duration = Duration::from_millis(5000);
```
ddd-mtl: Suggest renaming it to `PERSISTENCE_PERIODICITY_MS`  or `PERSISTENCE_PERIODICITY_MILLIS` as its more accurate and we know which time unit is used.

ddd-mtl:
As for the crate naming inquiry. I vote for `persistence_api` instead of `persistence`.
And instead of `persistence_impl` I would go with the name of the implementation:
- `persistence_mem`
- `persistence_pickle`

`persistence_mem` could actually be within the `persistence_api` crate if it doesn't have dependencies.

struktured:
> As for the crate naming inquiry. I vote for `persistence_api` instead of `persistence`.
> And instead of `persistence_impl` I would go with the name of the implementation:
> 
> * `persistence_mem`
> * `persistence_pickle`
> 
> `persistence_mem` could actually be within the `persistence_api` crate if it doesn't have dependencies.

I [renamed](ce3dfa5b7c2269d35a714874a98d35d20898327b) `persistence` to `persistence_api` and `persistence_impl` to_`persistence_[mem|pickle|file]` for the implementations.

crates/persistence_impl/src/eav/pickle.rs
struktured: > Suggest renaming it to PERSISTENCE_PERIODICITY_MS or PERSISTENCE_PERIODICITY_MILLIS as its more accurate and we know which time unit is used.
Done in fb44c36247adb3e0db790c7bf4fb686d95ed4e7e.

crates/persistence/README.md
```diff
@@ -0,0 +1,5 @@
+# lib3h_persistence
+
+A persistence api for lib3h and holochain. Provides abstractions for content addressable storage (CAS) and entity attribute value (index) associative stores.
+
+
```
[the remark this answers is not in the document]
struktured: > Please use the same README format as in the other published crates (lib3h, lib3h_protocol, etc.)

Done in 395ca68ffac4d730c634c512aad07c3c03292d18 and ce3dfa5b7c2269d35a714874a98d35d20898327b,

crates/persistence/src/error/mod.rs
struktured: > dead comment

Removed in 28de48bc064286a0cae8694053e33048ad4326c0.

crates/persistence_api/src/eav/eavi.rs
```diff
@@ -0,0 +1,493 @@
...
+        .is_ok());
+        assert!(EntityAttributeValueIndex::new(
+            &test_eav_entity().address(),
+            &ExampleAttribute::WithPayload("123".into()),
+            &test_eav_entity().address()
+        )
+        .is_ok());
+    }
+
+}
```
AshantiMutinta: I'm interested in the decision to make Attribute a trait, could you walk me through that?

crates/persistence_api/src/cas/content.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    /// the Content that would be stored in a ContentAddressableStorage
+    /// the default implementation covers anything that implements From<Foo> for JsonString
+    fn content(&self) -> Content;
+
+    /// restore/deserialize the original struct/type from serialized Content
+    /// the default implementation covers anything that implements From<JsonString> for Foo
+    fn try_from_content(content: &Content) -> Result<Self, PersistenceError>
+    where
+        Self: Sized;
```
dymayday: Can the content be a slice for example?

struktured: Open to improving the content api but in a subsequent PR.

crates/persistence_api/src/eav/eavi.rs
struktured: The attribute in holochain code was very specific to managing crud and links. The generalized form using an  attribute trait allows for other representations.

PR #105, opened 2019-06-12
Mirror DHT: First pass

ddd-mtl:
#### Refactored Dht trait & protocol
 - Use `EntryData` from Lib3hProtocol instead of 'data'
 - DhtEvent split with DhtCommand to separate orders by owner from notifications to send to the owner
e.g. `DataHoldRequest` renamed to `HoldEntry` command and `HoldEntryRequested` event.

#### Implemented MirrorDht
- Holds `PeerData` and `EntryData` in internal hashmap
- use rmp-serde to serialize gossip
- Only gossips on `HoldEntry` and `HoldPeer` to all other known nodes
- no timestamp for now

#### Dht tests
DHT semi-agnostic tests in dht/mod.rs

### Transport & P2pGateway
- advertise now returns an option since it needs a call to bind to be active
- renamed  P2pGateway field `transport` to `connection`

## Review checklist 


crates/lib3h_protocol/src/lib.rs
```diff
@@ -1,5 +1,10 @@
 //! This module provides the api definition for working with lib3h
 
+extern crate serde;
+#[macro_use]
+extern crate serde_derive;
+extern crate rmp_serde as rmps;
```
neonphog: I find it easier to figure out what's going on with original crate names, but I'm totally cool if others feel differently.

crates/lib3h/src/lib.rs
```diff
@@ -6,6 +6,15 @@ extern crate url;
 extern crate failure;
 #[macro_use]
 extern crate lazy_static;
+#[cfg(test)]
+#[macro_use]
+extern crate unwrap_to;
+extern crate serde;
+#[macro_use]
+extern crate serde_derive;
+extern crate rmp_serde as rmps;
```
neonphog: ditto rename

crates/lib3h/src/dht/mirror_dht.rs
```diff
@@ -0,0 +1,251 @@
...
+                if let Ok(entry) = maybe_entry {
+                    let is_new = self.add_entry(&entry);
+                    if is_new {
+                        return Ok(vec![DhtEvent::HoldEntryRequested(entry)]);
+                    }
+                    return Ok(vec![]);
+                }
+                de = Deserializer::new(&msg.bundle[..]);
+                let maybe_peer: Result<PeerData, rmp_serde::decode::Error> =
+                    Deserialize::deserialize(&mut de);
```
neonphog: For the real dht implementation, we will need to land a well defined gossip and p2p messaging protocol. I'm cool with this one being more organic. But I think we should strive for code readability and debug-ability.

It might be easier to see what's going on here if we have a super enum, and can always reliably encode / decode it:

```rust
pub enum Gossip {
  Entry(EntryData),
  Peer(PeerData),
}
```

ddd-mtl: yes, totally. Wanted to be sure I was going in the right direction before making this less sloppy :P

Issue #54, opened 2019-05-18
Implement Machine Ids


jmday:
As a holochain user, my machine has a unique identity on the transport network so that others can identify and connect to me.

- [x] lib3h transport layer can generate machineid base32 scheme (hcid) signing keypairs
- [x] lib3h can use my own hcid private key to sign data
- [x] lib3h can encode my own hcid pubkey so it can be advertised via my machine's uri
- [x] lib3h can use other machine_ids to validate other machines' signatures

Out of scope:
- persisting the private key

PR #111, opened 2019-06-18
Machine Id Support Code

neonphog:
RESOLVES https://github.com/holochain/lib3h/issues/54

## PR summary

This PR:

- introduces the `CryptoSystem` trait and associated sub-traits. It provides the beginning of a skeleton where we could potentially implement other crypto sub-systems
- causes `RealEngine` to accept a `CryptoSystem` trait generic and use it to generate a MachineId (aka TransportId).

## Review checklist 


crates/sodium/src/crypto_system/secure_buffer.rs
```diff
@@ -0,0 +1,130 @@
...
+            ProtectState::NoAccess => write!(f, "SecureBuffer( {:?} )", "<NO_ACCESS>"),
+            _ => write!(f, "SecureBuffer( {:?} )", *self),
+        }
+    }
+}
+
+impl std::ops::Deref for SecureBuffer {
+    type Target = [u8];
+
+    fn deref(&self) -> &[u8] {
```
dymayday: Same here `fn deref(&self) -> &Self::Target {` I believe

crates/sodium/src/crypto_system/secure_buffer.rs
```diff
@@ -0,0 +1,130 @@
...
+}
+
+impl BufferType for SecureBuffer {}
+
+impl Buffer for SecureBuffer {
+    fn new(size: usize) -> CryptoResult<Self> {
+        check_init();
+        let z = unsafe {
+            // round up to nearest 8
+            let align_size = (size + 7) & !7;
```
dymayday: What's this magic doing? ^^'

dymayday:
I don't have enough knowledge for now about crypto for checking the logic behind the impl tho :/

crates/sodium/src/crypto_system/secure_buffer.rs
neonphog: sodium_malloc has some undefined behavior if you give it a size that is not memory-aligned (https://download.libsodium.org/doc/memory_management#guarded-heap-allocations), so we need to round up to the nearest 8 bytes.

crates/crypto_api/src/buffer/insecure_buffer.rs
```diff
@@ -0,0 +1,56 @@
+use crate::{Buffer, BufferType, CryptoResult, ProtectState};
+
+/// You probably just want to use Vec<u8> directly rather than this.
+/// This is a class is mainly an implementation reference for SecureBuffers.
+#[derive(Debug, Clone)]
+pub struct InsecureBuffer {
+    b: Box<[u8]>,
```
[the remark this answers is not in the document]
neonphog: This is mimicking the behavior of [SecureBuffer](https://github.com/holochain/lib3h/pull/111/files#diff-a391cd153809f9242fb562af4b305f5bR70) which is not dynamically sizable, I also implement the [Buffer trait on Vec<u8>](https://github.com/holochain/lib3h/pull/111/files#diff-17256567e3d16eeb4f1d81d3e54ca79cR80) for more general usage.

crates/crypto_api/src/buffer/insecure_buffer.rs
```diff
@@ -0,0 +1,56 @@
...
+    p: std::cell::RefCell<ProtectState>,
+}
+
+impl std::ops::Deref for InsecureBuffer {
+    type Target = [u8];
+
+    fn deref(&self) -> &[u8] {
```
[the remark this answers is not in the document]
neonphog: Good call :+1: - updated

crates/sodium/src/crypto_system/secure_buffer.rs
neonphog: Good call :+1: - updated

crates/lib3h/src/real_engine.rs
```diff
@@ -41,11 +43,25 @@ pub struct RealEngine<T: Transport> {
...
     /// Map of P2p gateway per Space+Agent
     space_gateway_map: HashMap<PlayerId, P2pGateway<TransportSpace, RrDht>>,
+    /// Our TransportId (aka MachineId, "HcMyadayada")
+    _transport_id: String,
+    /// The TransportId public key
+    _transport_public_key: Vec<u8>,
+    /// The TransportId secret key
+    _transport_secret_key: SecBuf,
+    /// needed to accept the Crypto trait generic
+    _phantom_crypto: std::marker::PhantomData<Crypto>,
```
struktured: Why the '_' prefixes here? Are these unused fields?

crates/lib3h/src/real_engine.rs
```diff
@@ -41,11 +43,25 @@ pub struct RealEngine<T: Transport> {
...
 }
 
-impl RealEngine<TransportWss<std::net::TcpStream>> {
+impl<SecBuf: Buffer, Crypto: CryptoSystem>
+    RealEngine<TransportWss<std::net::TcpStream>, SecBuf, Crypto>
+{
     /// Constructor
     pub fn new(config: RealEngineConfig, name: &str) -> Lib3hResult<Self> {
```
struktured: Since this function returns a result, it would be better from an error handling perspective to keep this code entirely in a result monad. A `map_err` might be required to make the return type compatible but that's probably it. Now if you determined these local results are panic level problems that have no value in being propagated outside this function as a return value then the `?` is fine, but note here that the function never returns a `Err(..)` as currently written, which seems suspect.

crates/lib3h/src/real_engine.rs
neonphog: Since the struct doesn't have any members that use the `Crytpo` trait (used in member functions), we need to have the phantom data marker to satisfy the compiler... but it's not really there, hence the `_`.

The others are temporarily `_` to satisfy the linter until we implement code that uses them.

crates/lib3h/src/real_engine.rs
neonphog: I just hate writing code in functions that don't return results... it's so cluttered and harder to read with all the matches / error mapping. `?`s are so nice.

That said, there's probably some refactoring that could be done here, especially since there are two constructors in this file with duplicate logic for machineId generation.

crates/lib3h/src/real_engine.rs
ddd-mtl: Using `_` on struct fields is quite broken and painful to use.
As an alternative, using `#[allow(dead_code)]` also makes the linter happy.

crates/lib3h/src/real_engine.rs
ddd-mtl: @struktured The function doesn't return an `Err(...)` right now because we haven't finished implementing it. I except it to return an error if the config argument is bad ; once we have actual config handling.

crates/lib3h/src/engine/mod.rs
```diff
@@ -46,4 +47,12 @@ pub struct RealEngine<T: Transport, D: Dht> {
...
     /// Map of P2p gateway per Space+Agent
     space_gateway_map: HashMap<ChainId, P2pGateway<P2pGateway<T, D>, D>>,
+    /// Our TransportId (aka MachineId, "HcMyadayada")
+    _transport_id: String,
+    /// The TransportId public key
+    _transport_public_key: Vec<u8>,
+    /// The TransportId secret key
+    _transport_secret_key: SecBuf,
+    /// needed to accept the Crypto trait generic
+    _phantom_crypto: std::marker::PhantomData<Crypto>,
```
ddd-mtl: Maybe group those fields in a struct?

crates/lib3h/src/real_engine.rs
zippy: I'm only ok with `allow(dead_code)` if it won't bleed over and allow other bits of dead code to sneak in.

crates/crypto_api/src/lib.rs
```diff
@@ -1,9 +1,54 @@
 //! lib3h abstract cryptography traits and data types
+//!
+//! # Examples
+//!
+//! ```
+//! extern crate lib3h_crypto_api;
+//!
+//! // It is recommended to use CryptoSystem as a Generic Trait like this:
```
zippy:
```suggestion
//! // CryptoSystem is designed to be used as a Generic Trait like this:
```

crates/sodium/src/crypto_system/mod.rs
```diff
@@ -0,0 +1,44 @@
...
+/// # Examples
+///
+/// ```
+/// extern crate lib3h_crypto_api;
+/// use lib3h_crypto_api::Buffer;
+///
+/// extern crate lib3h_sodium;
+/// use lib3h_sodium::SecureBuffer;
+///
+/// // It is recommended to use CryptoSystem as a Generic Trait like this:
```
zippy:
```suggestion
/// // CryptoSystem is designed to be used as a Generic Trait like this:
```

crates/sodium/src/crypto_system/secure_buffer.rs
```diff
@@ -0,0 +1,131 @@
...
+    fn drop(&mut self) {
+        unsafe {
+            rust_sodium_sys::sodium_free(self.z);
+        }
+    }
+}
+
+impl Clone for SecureBuffer {
+    fn clone(&self) -> Self {
+        let mut out = SecureBuffer::new(self.s).expect("could not alloc new");
```
zippy: This looks like a runtime problem.  Shouldn't we not be doing an `expect` here?

neonphog: Alas, clone must return Self, not a result, so we have to panic!... But rust panics on all memory allocation errors, so it's consistent at least.

crates/lib3h/src/engine/mod.rs
neonphog: grouped :+1:

crates/lib3h/src/real_engine.rs
neonphog: yeah, when no longer dead we need to be careful to remove it... and also, make sure it's as local as possible.

I went ahead and switched to it for now.

PR #119, opened 2019-06-25
Lib3h Mock: Two-way SendDirectMessage

ddd-mtl:
## PR summary
Have a Peer gossip back its PeerData when receiving HoldPeer on the network gateway so nodes can talk to each other after a connection.
Added and implemented P2pProtocol::AllJoinedSpaceList so nodes can join spaces first before connecting.

### Misc
Renamed reverse_map to connection_map.
Renamed P2pProtocol::JoinSpace  to P2pProtocol::BroadcastJoinSpace.

## Review checklist 


crates/lib3h/src/gateway/gateway_dht.rs
```diff
@@ -102,7 +102,7 @@ impl<T: Transport, D: Dht> P2pGateway<T, D> {
                     // If no connection to that transportId is open, open one first.
                     self.inner_transport
                         .borrow_mut()
-                        .send(&[peer_transport.as_str()], &payload)?;
+                        .send(&[peer_transport.path()], &payload)?;
```
struktured: This should be wrapped in the `transport::url_to_transport_id` function so we can easily isolate the subtle assumptions later. I will do it in a future PR.

PR #118, opened 2019-06-24
Implement server side transport trait for tcp/tls

struktured:
## PR summary

This PR implements a transport trait for web sockets, with emphasis on fleshing out the server side bind and accept implementation, as well as the necessary TLS hand shaking.

## Highlights
 
 - introduced functional approach to associating a bind implementation to a transport
 - implemented a bind for tcp transports
 - implemented server side tls hand shaking
 - changed transport address to `url::Url` for clarity and type safety.

## Review checklist 
- [x] The story has unit or integration tests
   See the `transport::tests::wss_send_tests`

- [x] No new bugs, and any tech-debt is identified and justified
  - Add several TODOs related to url manipulation within transports
- [ ] There is enough API documentation (how to use)
- [x] There is enough code documentation (how the code works)
   - Documented `Bind` and `Acceptor` function signatures

PR #123, opened 2019-06-27
Transport websocket send test

struktured:
## PR summary

This PR fixes several remaining problems from  #118:

 - sets `non_blocking` to true on the tcp acceptor socket (fixes blocking issue)
 - adds a static counter to both the memory and websocket id generator to reduce id collisions between transports
 - generalizes the `transport::tests::send_test` to support websockets
   - process loops are executed a constant number of times rather than just once to complete the tls hand shaking
   - both nodes A and B are now processed in the loop rather than just one or the other
   - the uri equal to connection id assertions instead now assert existance of a uri given a connection id but no longer verify exact values (might require some discussion)   
 - to support similar `transport::get_uri` functionality in the memory trait that the websocket trait already has, the memory trait's `get_uri` function now checks if its bound server has the connection id- if it does it returns its own uri.
 - uncomment out the `wss_send_test`, which now passes

## Review checklist 
  - uncommented out the transport `wss_send_test` which tests a sending of a payload to and from a pair of nodes
  - need to discuss semantics of get_uri again to make sure it's legitimate


crates/lib3h/src/transport/mod.rs
```diff
@@ -64,56 +66,61 @@ pub mod tests {
...
-        println!("calling node A process 3");
-        let (did_work, event_list) = node_A.process().unwrap();
+        did_work = false;
+        event_list.clear();
+        for _x in 0..NUM_PROCESS_LOOPS {
+            let (did_work_A, mut event_list_A) = node_A.process().unwrap();
+            let (_did_work_B, _event_list_B) = node_B.process().unwrap();
+            did_work |= did_work_A;
+            event_list.append(&mut event_list_A);
+        }
```
ddd-mtl: It might be best to have infinite loop with a break condition and a timeout panic.

ddd-mtl:
and thanks for adding the counter to TransportMemory! 🙏

crates/lib3h/src/transport/memory_mock/memory_server.rs
```diff
@@ -66,9 +67,14 @@ impl MemoryServer {
             uri: uri.clone(),
             inbox_map: HashMap::new(),
             new_conn_inbox: Vec::new(),
+            connection_ids: HashSet::new(),
         }
     }
 
+    pub fn is_connection(&self, id: &ConnectionIdRef) -> bool {
```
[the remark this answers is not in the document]
struktured: I changed it to `has_connection`.

crates/lib3h/src/transport/mod.rs
struktured: I made a TODO about this for now right above the code block.

PR #125, opened 2019-06-27
nixify

thedavidmeister:
## Review checklist 


neonphog:
this is still WIP, so maybe I'm missing something, but nix-shell never returns for me... stuck at:

```
building '/nix/store/x56qfhphc9qb42n8f0v3afy3xbp8wra2-nixpkgs-channels-8634c3b.drv'...
exporting https://github.com/NixOs/nixpkgs-channels.git (rev 8634c3b619909db7fc747faf8c03592986626e21) into /nix/store/5m8hfg0zcxl5rv25yqlap6cjbhf3y0mv-nixpkgs-channels-8634c3b
Initialized empty Git repository in /nix/store/5m8hfg0zcxl5rv25yqlap6cjbhf3y0mv-nixpkgs-channels-8634c3b/.git/
remote: Enumerating objects: 163, done.        
remote: Counting objects: 100% (163/163), done.        
remote: Compressing objects: 100% (130/130), done.        
```

neonphog:
@thedavidmeister - would it be ok if we keep the current Makefile targets (perhaps not as the default)?

I like having everything on my base system, it's so much faster than going into the nix shell...

I could do something like:

```
make native-test
```

? I'm happy to personally maintain those targets.

struktured:
@neonphog it completed for me but took over 20 minutes at least. The 2nd run only took 14 seconds though.

thedavidmeister:
@neonphog @struktured the normal process is that you would leave the nix shell open rather than redoing it for every command, makefile is not really supported standalone, i threw in a wrapper for backwards compatibility rather than it being the best way to work

neonphog:
@thedavidmeister - I absolutely love the nix shell from a CI perspective ESPECIALLY because it lets me easily run the CI process locally to verify things are going to work before I push code and / or in case I need to debug something CI-specific. Thank you so much for all the hard work you've put into it!

And. Developers are a very particular bunch. I've got my base system set up the way I want it. I find the nix shell restrictive and foreign. I do not want to be stuck in it while I'm iterating on development. I'm going to maintain a Makefile for my own personal use. Are you suggesting I shouldn't maintain my Makefile targets as part of the code repo? What if others feel as I do about the shell they're using?

thedavidmeister:
@neonphog cc: @pjkundert @zippy 

we discussed this on zoom and i'm putting this summary out for wider visibility

i know where you are coming from and agree
you understand where i am at as well and agree

so what does that mean?

**if what you're doing locally isn't a source of decoherence then iSquid don't care**

we're trying _really_ hard to respect and encourage diversity of local development environments
BUT...
the moment "your machine" touches "my machine" in a divergent way then iSquid cares a lot!
AND...
iSquid wants people to have _awesome_ tools to work in a convergent way when that is important

agent-centricity applied to ops!

in pragmatic and general terms not at all specific to you @neonphog, everyone can go nuts on their own makefile(s) _provided_:

- we keep a main/default makefile that wraps the standard nix commands as this is a clear integration point for external make based tooling/consumption, it's likely that we'll expand/build on this pattern over time
- any personal makefiles are named/documented and only kick around as long as they are maintained
- we target nix commands in CI/releases to hit all the weird RUSTFLAGS, compiler versions, etc. that we track for downstream compatibility
- myself and others like @samrose et al can only provide limited support for bespoke commands, and our standard troubleshooting approach will often start (and possibly end) at "try the standard commands"
- if you have specific frustrations with any tooling at all please raise them upstream so we have an opportunity to improve based on real world usage/feedback
- nobody ever _needs_ something from a personal makefile to participate
- nobody is ever "blocked" pending a personal makefile refactor

which means, @neonphog if you can resurrect the old makefile how you'd like it under your name or something, that should mean we can land all of this i think, which is exciting

thedavidmeister:
@neonphog yeah cool, happy to chat re: the opt level, i think that's there mostly for historical reasons

maybe start a thread in holonix issue queue?

neonphog:
> maybe start a thread in holonix issue queue?

done :+1:

neonphog:
fyi, @thedavidmeister - the CI failure here looks like another upstream rust dependency debacle:

```
error[E0432]: unresolved import `core::task::LocalWaker`
 --> /root/project/.cargo/registry/src/github.com-1ecc6299db9ec823/futures-core-preview-0.3.0-alpha.12/src/future/future_obj.rs:6:12
  |
6 |     task::{LocalWaker, Poll},
  |            ^^^^^^^^^^ no `LocalWaker` in `task`
```

they say this is fixed in futures-core-preview-0.3.0-alpha.13, but we're including deps that are still depending on 12 : (

samrose:
It seems in theory that you could develop in any way as long as your code can build and test in holonix.

PR #135, opened 2019-06-29
NodeMock

ddd-mtl:
## PR summary
Imported and adapted the `TestNode` from `test_bin`. Renamed it `NodeMock.`
Renamed what was the former `integration_test.rs`  to  `engine_test.rs` (no actual change in engine_test.rs)
Now `integration_test.rs` is for testing with NodeMocks.
In this PR, it currently has the same `send_direct_message` test as in engine_test.
You can see how easier it is to write tests with that.

## Review checklist 


crates/lib3h/tests/utils/predicate.rs
```diff
@@ -0,0 +1,31 @@
+/// Check if Protocol is of type $p
+#[allow(unused_macros)]
+macro_rules! one_is {
```
zippy: recommend renaming this from "one_is" to "it_is_a"

crates/lib3h/tests/utils/predicate.rs
```diff
@@ -0,0 +1,31 @@
...
+                return true;
+            }
+            return false;
+        }
+    };
+}
+
+/// Check if Protocol is of type $p and meets conditions set in $code
+#[allow(unused_macros)]
+macro_rules! one_is_where {
```
zippy: similarly "it_is_a_where"

crates/lib3h/tests/utils/predicate.rs
```diff
@@ -0,0 +1,31 @@
...
+    ($p:pat, $code:tt) => {
+        move |d| return if let $p = d { $code } else { false }
+    };
+}
+
+#[allow(unused_macros)]
+macro_rules! one_let {
```
zippy: is this used anywhere?

ddd-mtl: Not yet here. It was used a lot in `test_bin`

PR #137, opened 2019-07-01
RFC: first cut at capnproto p2p + transport encoding protocol schemas

neonphog:
## PR summary

This looks like 5,055 line additions, but most of that is generated code you don't have to review. There are actually only 529 lines of code added.

Added:
- capnproto schema for p2p_protocol
- capnproto schema for transit_encoding protocol
- capnp_build tool crate for compiling the schemas into generated rust code

See capnp_build/README.md for notes on capnp_build, but basically we're committing the generated code to the repo, rather than trying to do it in a build.rs on the fly because it's unreasonable to expect every lib3h or holochain developer to have the capnproto compile tool installed and in their path.

## Review checklist 


neonphog:
blarg, n/m. I thought I'd figured out a way to make the channels work like this after all, but I was wrong. I *do* need to split this up into 3 protocols.

crates/p2p_protocol/protocol/p2p.capnp
```diff
@@ -0,0 +1,179 @@
...
+    aspectConstraintTime @1 :AspectConstraintTime;
+    # constrain aspectHashList by publish timestamp
+
+    aspectConstraintCount @2 :AspectConstraintCount;
+    # constrain aspectHashList by local store count
+
+    aspectHashList @3 :List(AspectHashList);
+    # list of aspectHashes associated with entryAddresses
+    # that fall within all the above constraints.
+  }
```
ddd-mtl: Suggest renaming `aspectHashList` to `aspectHashListResponse` or something of this sort to make clear that the field is used for the response only.

crates/p2p_protocol/protocol/p2p.capnp
```diff
@@ -0,0 +1,179 @@
...
+
+  struct AspectHashData {
+    aspectHash @0 :Data;
+    # the hash of the aspectData
+
+    aspectData @1 :Data;
+    # the data associated with the aspectHash
+  }
+
+  struct AspectConstraintLoc {
```
ddd-mtl: Suggest renaming it to `AspectConstraintArc` since this struct is actually defining an Arc of Locs and  an "Arc" is always an Arc of Locs, right?

crates/p2p_protocol/protocol/p2p.capnp
```diff
@@ -0,0 +1,179 @@
...
+    locArcStart @0 :UInt32;
```
ddd-mtl: `arcStart`

crates/p2p_protocol/protocol/p2p.capnp
```diff
@@ -0,0 +1,179 @@
...
+
+  struct AspectHashDataList {
+    entryAddress @0 :Data;
+    # when asking for aspect hash data, we also want to know the entryAddress
+
+    aspectHashDataList @1 :List(AspectHashData);
+    # the list of pairs of aspectHash / aspectData
+  }
+
+  struct AspectHashData {
```
ddd-mtl: `AspectData`

crates/p2p_protocol/protocol/p2p.capnp
```diff
@@ -0,0 +1,179 @@
...
+    # associated with an entryAddress
+
+    aspectHashList @1 :List(Data);
+    # the list of aspectHashes associated with the above entryAddress
+    # note this list is probably not comprehensive,
+    # it may be only those that fall within constraints, or only
+    # those that are being requested / responded with / etc.
+  }
+
+  struct AspectHashDataList {
```
ddd-mtl: Suggest renaming to `AspectDataList` to make the difference with `AspectHashList` more obvious.

crates/p2p_protocol/protocol/p2p.capnp
```diff
@@ -0,0 +1,179 @@
...
+
+    aspectData @1 :Data;
+    # the data associated with the aspectHash
+  }
+
+  struct AspectConstraintLoc {
+    locArcStart @0 :UInt32;
+    # the start location of the rrdht arc (inclusive)
+
+    locArcEnd @1 :UInt32;
```
ddd-mtl: `arcEnd`

crates/p2p_protocol/protocol/p2p.capnp
zippy: In the case of AspectHashData, storing the hash in the tuple probably means we need to calculate it as well to verify it, and if so why even store it?

crates/p2p_protocol/protocol/transit_encoding.capnp
```diff
@@ -0,0 +1,153 @@
...
+
+  enum ReasonCode {
+    unspecified @0;
+    badMagic @1;
+    badEncoding @2;
+    badNetworkId @3;
+    kxDecodeFail @4;
+    unexpectedSigPubKey @5;
+    badSignature @6;
+    encDecodeFail @7;
```
zippy: Assuming it's easy to expand reason codes in the future.  But for the moment makes things brittle?  Would it make sense to allow text into the reason?  Or something app-dependant, like "banned" or "blocked by warrant< hash>".

crates/p2p_protocol/protocol/p2p.capnp
```diff
@@ -0,0 +1,179 @@
...
+    # text indicating details of error
+
+    enum ErrorCode {
+      unknown @0;
+      # default if error is not well-known, or if remote is using a newer proto
+    }
+  }
+
+  struct MsgHandshake {
+    fromStoreArc @0 :UInt32;
```
ddd-mtl: I don't get what the `from` is suppose to mean here.
Is this value the starting loc of the Arc or the Arc length?
If its the starting loc, I'd suggest renaming to `storeArcStart`
If its the Arc length I'd suggest renaming to `storeArcLength`
Also, is the length the full breadth of the arc or the half length since its suppose to be symmetric from the agent loc?

crates/p2p_protocol/protocol/p2p.capnp
```diff
@@ -0,0 +1,179 @@
...
+    # the storage arc length of the source transport / agent
+
+    fromQueryArc @1 :UInt32;
```
ddd-mtl: same here

crates/p2p_protocol/protocol/transit_encoding.capnp
```diff
@@ -0,0 +1,153 @@
...
+
+    sodiumJson @3;
+    # libsodium-encrypted Json encoded
+
+    sodiumPacked @4;
+    # libsodium-encrypted packed Capnproto encoded
+  }
+}
+
+struct MsgLsnH2 {
```
zippy: Is there a reason why these struct names need to be so short?  Usually I'm ok with that, but `LsnH2` isn't transparent in meaning to me?

crates/p2p_protocol/protocol/p2p.capnp
```diff
@@ -0,0 +1,179 @@
+# lib3h p2p wire protocol
+# note, initial handshake will use transit-enc.capnp
+# once that is complete, messages from this schema
+# will be encoded as negotiated.
+
+@0x859198991b95d3e1;
+
+struct Message {
```
ddd-mtl: Suggest renaming to `P2pMessage` as `Message` sounds too generic and also collisions with `Message` in `transport_dm`.

crates/p2p_protocol/protocol/transport_dm.capnp
```diff
@@ -0,0 +1,100 @@
+# For Holochain, the app-level connections use DirectMessage for node2node
+# communications between app developer code.
+# At the transport-level, once a connection is established between two nodes,
+# We use DirectMessage within lib3h to set up multiplexing channels,
+# and negotiate relay services, etc.
+
+@0xdd5a14cf7a6734dc;
+
+struct Message {
```
ddd-mtl: Suggest renaming to `TransportMessage` as `Message` sounds too generic and also collisions with `Message` in `p2p.capnp`.

crates/p2p_protocol/protocol/transit_encoding.capnp
ddd-mtl: +1

crates/p2p_protocol/protocol/p2p.capnp
neonphog: The "from" is supposed to indicate it's the arc of the node sending the message. Any ideas for something that would be more clear?

Yep, it's the half length. `0` is no arc, `1` means covering exactly the location that their agentId hashes to (let's say 42), `2` would indicate anything (inclusively) between 41 and 43.

`sourceAgentStoreArcHalfLength`?

crates/p2p_protocol/protocol/p2p.capnp
neonphog: Actually this is used in both directions. The node making the request sends every aspectHash within the constraints, then the responding node sends only those hashes that both fall within the constraints AND the requesting node didn't include in the request. (because they already have those).

neonphog: Then both nodes individually request data for all the hashes they don't already know about.

crates/p2p_protocol/protocol/p2p.capnp
neonphog: I agree we need to validate it if included.

It would simplify the api, and lower bandwidth to remove it.

Part of me likes the explicitness of having the hash in there, like, if a node had corrupted aspect data on its disk and published it, we could reject that at the protocol layer before even passing it off to core for validation. Buuut, if core is going to validate it anyway, perhaps we don't want to double that up?

crates/p2p_protocol/protocol/transit_encoding.capnp
neonphog: Yep! there's the reasonText above, so if the error isn't a well-known code, the code can be unspecified and the error can be enumerated in text.

We may want to remove some of these codes... really, well-known codes are only useful if they are something that can be recovered from in an automated code path. Like the `WouldBlock` error sent back from a non-blocking tcp session that means, we don't have data right now, try back later.

crates/p2p_protocol/protocol/transit_encoding.capnp
neonphog: hehe : ) Yeah, I can expand these a bit. Basically, one set of messages is coming from the connecting node ("Con") and the other set is coming from the listening node ("Lsn"), they are all numbered handshake messages, so handshake step 1 ("H1"), handshake step 2 ("H2"), etc.

crates/p2p_protocol/protocol/p2p.capnp
ddd-mtl: I don't see how it can be the arc of someone else than the node sending the message, so I don't find this info relevant. Maybe it would if we had `fromStoreArc` and `toStoreArc` in that struct, but thats not the case so I'd go with `storeArcHalfLength` or `storeRadius` ;)

crates/lib3h_protocol/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "0.0.3-alpha1"
 authors = ["Holochain Core Dev Team <devcore@holochain.org>"]
 edition = "2018"
 # crates.io stuff
-description = "Lib3h Protocol definition for P2P communication."
+description = "Lib3h Protocol definition for API interaction with the library."
```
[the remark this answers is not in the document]
neonphog: :+1:

crates/p2p_protocol/protocol/p2p.capnp
neonphog: :+1:

crates/p2p_protocol/protocol/p2p.capnp
neonphog: :+1:

crates/p2p_protocol/protocol/p2p.capnp
neonphog: I'll go ahead and remove it for now, we can evaluate if that keeps making sense as we go forward.

crates/p2p_protocol/protocol/p2p.capnp
neonphog: :+1:

crates/p2p_protocol/protocol/p2p.capnp
neonphog: :+1:

crates/p2p_protocol/protocol/p2p.capnp
neonphog: :+1:

crates/p2p_protocol/protocol/transport_dm.capnp
neonphog: I think I'm going to rename this whole file to `multiplex.capnp` then, this struct name can be `MultiplexMessage` :+1:

ddd-mtl: 👍

crates/p2p_protocol/protocol/p2p.capnp
ddd-mtl: Oh okay, sweet! Forget my initial comment then.

crates/p2p_protocol/protocol/transit_encoding.capnp
neonphog: Alright, flipped this around and expanded: `MsgStep1FromConnect`, `MsgStep2FromListen`, etc

neonphog:
@zippy RE channel state

> I just want confirm that this is a reasonable tradeoff at this point, because it means we need to handle and test all those cases of extra variable state on either side. A side question: should there be a way to send a message without having created the channel first, i.e. with the full header info?

Given the spaceHashes and agentIds are all 32 bytes long, this brings the per-message overhead for this channel info down from 96 bytes to 4. Obviously long term (especially in a micro-service environment with lots of apps using lots of channels to communicate) I think this is very worth it.

From a code usage perspective, this should be abstracted away, so you'd still send messages with spaceHash, toAgent, and fromAgent included, then in the multiplexer module it would detect if it had created a channel for that combo yet, and send the ChannelCreate message if needed, so I don't think there's a need for an alternate api that includes these inline. (it would only save us 8 bytes for the channelId on both messages anyways).

For the short term, I don't have a super strong opinion about skipping this or not. I don't feel like the test cases to cover channelId mismatch glitches will be too onerous, but that could just be me being optimistic : )

crates/p2p_protocol/protocol/p2p.capnp
neonphog: What do you think about `storeArcRadius` and `queryArcRadius` just to be a little extra explicit?

ddd-mtl: I think its nice 👍

ddd-mtl:
+1 on having channels. Dealing with space+agent is already a pain so having channels may add a bit of complexity but can also help in streamlining some parts.

neonphog:
In-Transit Sequence Diagram

```sequence
Note right of Alice: Step 1: Open a connection\nsending config info\nand our kx pubkey
Alice->Bob: MsgStep1FromConnect
Note right of Bob: Bob can drop if we're\nworried about DDoS,\nor respond with\nHalt or Continue
Note right of Alice: Step 2: listening node's\nkx pubkey, sig pubkey,\nplus l2c session key
Bob->Alice: MsgStep2FromListen
Note right of Alice: Step 3: connect node's\nsig pubkey, c2l session key,\nplus signature
Alice->Bob: MsgStep3FromConnect
Note right of Bob: Bob verifies that Alice\nsigned his session key\nwith her sig private key
Note right of Alice: Step 4: Bob responds with\nhis signature
Bob->Alice: MsgStep4FromListen
Note right of Alice: Step 5: Accept Signature
Alice->Bob: MsgStep5FromConnect
Note right of Bob: Now that Alice has\naccepted Bob's signature,\nwe can continue exchanging\nencrypted messages
```

in-transit-sequence

neonphog:
- [P2p Sequence Diagram](https://github.com/holochain/lib3h/blob/p2p-protocol/crates/p2p_protocol/doc/p2p_protocol.md)

Issue #126, opened 2019-06-27
in-memory mock: list (get-hold and get publish) for startup initialization


zippy:
Acceptance:

test of startup

PR #145, opened 2019-07-04
Mock: Get*Lists

ddd-mtl:
## PR summary
Solves #126

Send Get*Lists requests on JoinSpace.
On receiving authoring-list, ask Core for each entry and broadcast each response.
On receiving gossiping-list, send a HoldEntryAddress command to the DHT for each entry.

Added Get*lists test suite. Imported from `test_bin`.

### Follow-up

Handling  multiple aspects per entry will be done in follow-up PRs.

### PR Details

Added the test function's name to a node's binding url when testing in-memory transport.
Otherwise we would get conflicts because `cargo test` will run test suites in parallel and the in-memory transport server can't manage multiple clients with the same url.

Refactored testing succesful get of `ENTRY_ADDRESS_1` from a node as the function `request_entry_1()`, because it is used by multiple tests.

## Review checklist 

### DIFF
--- a/crates/lib3h/src/engine/real_engine.rs
+++ b/crates/lib3h/src/engine/real_engine.rs
@@ -168,7 +168,7 @@ impl<T: Transport, D: Dht, SecBuf: Buffer, Crypto: CryptoSystem> NetworkEngine
         // Process the space layer
         let mut p2p_output = self.process_space_gateways()?;
         outbox.append(&mut p2p_output);
-        trace!("{} - RealEngine.process() END\n", self.name);
+        trace!("RealEngine.process() END - (outbox: {})\n", outbox.len());
         // Done
         Ok((inbox_did_work || net_did_work, outbox))
     }
@@ -218,8 +218,8 @@ impl<T: Transport, D: Dht, SecBuf: Buffer, Crypto: CryptoSystem> RealEngine<T, D
                 Transport::post(&mut *self.network_gateway.borrow_mut(), cmd)?;
             }
             Lib3hClientProtocol::JoinSpace(msg) => {
-                let output = self.serve_JoinSpace(&msg)?;
-                outbox.push(output);
+                let mut output = self.serve_JoinSpace(&msg)?;
+                outbox.append(&mut output);
             }
             Lib3hClientProtocol::LeaveSpace(_msg) => {
                 // FIXME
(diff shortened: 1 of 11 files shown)


crates/lib3h/src/engine/real_engine.rs
```diff
@@ -284,24 +284,32 @@ impl<T: Transport, D: Dht, SecBuf: Buffer, Crypto: CryptoSystem> RealEngine<T, D
...
                 match maybe_space {
                     Err(res) => outbox.push(res),
                     Ok(space_gateway) => {
-                        let response = FetchDhtEntryResponseData {
-                            msg_id: msg.request_id.clone(),
-                            entry: msg.entry.clone(),
-                        };
-                        let cmd = DhtCommand::EntryDataResponse(response);
-                        space_gateway.post_dht(cmd)?;
+                        if msg.request_id == "__author_list" {
```
neonphog: : ) - I think we might need to create a rust equivalent of https://github.com/holochain/n3h/blob/master/lib/n3h-common/track.js at some point.

crates/lib3h/src/lib.rs
```diff
@@ -31,7 +31,9 @@ pub mod tests {
     // for this to actually show log entries you also have to run the tests like this:
     // RUST_LOG=lib3h=debug cargo test -- --nocapture
     pub fn enable_logging_for_test(enable: bool) {
-        //std::env::set_var("RUST_LOG", "debug");
+        if std::env::var("RUST_LOG").is_err() {
+            std::env::set_var("RUST_LOG", "debug");
+        }
```
neonphog: :+1:

crates/lib3h/src/engine/real_engine.rs
```diff
@@ -465,10 +513,25 @@ impl<T: Transport, D: Dht, SecBuf: Buffer, Crypto: CryptoSystem> RealEngine<T, D
...
+            request_id: "gossiping".to_owned(),
+        };
+        output.push(Lib3hServerProtocol::HandleGetGossipingEntryList(
+            list_data.clone(),
+        ));
+        list_data.request_id = "authoring".to_owned();
+        output.push(Lib3hServerProtocol::HandleGetAuthoringEntryList(list_data));
         // Done
-        Ok(Lib3hServerProtocol::SuccessResult(res))
+        output.push(Lib3hServerProtocol::SuccessResult(res));
```
neonphog: I realize it shouldn't actually matter, but do you think we should respond with the success to the JoinSpace request first, before making the list requests?

crates/lib3h/src/engine/real_engine.rs
ddd-mtl: Yeah­. I'll add a todo for that.

crates/lib3h/src/engine/real_engine.rs
ddd-mtl: Good point

Issue #146, opened 2019-07-04
in-memory mock: Handle multiple aspects


ddd-mtl:
Figure out how we want tracking of aspects done, and add multiple aspects tests in test suites.

Open questions:
 - Should DHT double check that an entry and all its aspects are already known and should not be broadcasted even when commanded to?
e.g. Should lib3h respond with a FailureResult if Core sends the same PublishEntry twice?

-> PublishEntry should respond with CheckHealth for each aspect. To clarify when doing CheckEntryHealth

PR #148, opened 2019-07-05
Mock: handle multiple aspects

ddd-mtl:
## PR summary
Solves #146

Changed: 
`DhtCommand::HoldEntryAddress(Address)` to:
`DhtCommand::HoldEntryAspectAddress(EntryData)`
since hold commands should be per aspect.

Added `get_aspects_of(entry_address)` to dht_trait.

DHT implementations should have an internal diff function to figure out if aspects are new or not.

Added some multiple aspects tests, imported from `test_bin`
Made `EntryAspectData` sortable for deterministic `EntryData` comparaisons.

#### Open question

Maybe create a type like this:
`type AspectAddressList = HashSet<Address>`
And use that in `GetListData` and `FetchEntryData`?

## Review checklist 

### DIFF
--- a/crates/lib3h/src/engine/real_engine.rs
+++ b/crates/lib3h/src/engine/real_engine.rs
@@ -455,6 +455,8 @@ impl<T: Transport, D: Dht, SecBuf: Buffer, Crypto: CryptoSystem> RealEngine<T, D
                                 };
                                 aspect_list.push(fake_aspect);
                             }
+                            // Create "fake" entry, in the sense an entry with no actual content,
+                            // but valid addresses.
                             let fake_entry = EntryData {
                                 entry_address: entry_address.clone(),
                                 aspect_list,
--- a/crates/lib3h/src/engine/real_engine.rs
+++ b/crates/lib3h/src/engine/real_engine.rs
@@ -341,7 +341,7 @@ impl<T: Transport, D: Dht, SecBuf: Buffer, Crypto: CryptoSystem> RealEngine<T, D
                 match maybe_space {
                     Err(res) => outbox.push(res),
                     Ok(space_gateway) => {
-                        let cmd = DhtCommand::HoldEntryAddress(msg.entry.entry_address);
+                        let cmd = DhtCommand::HoldEntryAspectAddress(msg.entry);
                         space_gateway.post_dht(cmd)?;
                         // Dht::post(&mut space_gateway, cmd)?;
                     }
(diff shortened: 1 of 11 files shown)


crates/lib3h/src/engine/real_engine.rs
```diff
@@ -561,3 +581,11 @@ impl<T: Transport, D: Dht, SecBuf: Buffer, Crypto: CryptoSystem> RealEngine<T, D
         Err(Lib3hServerProtocol::FailureResult(res))
     }
 }
+
+/// Return true if all elements of list_b are found in list_a
+fn includes(list_a: &[Address], list_b: &[Address]) -> bool {
+    let set_a: HashSet<_> = list_a.iter().map(|addr| addr).collect();
+    let set_b: HashSet<_> = list_b.iter().map(|addr| addr).collect();
+    let diff: HashSet<_> = set_a.difference(&set_b).map(|addr| addr).collect();
```
neonphog: https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.is_subset

crates/lib3h/src/engine/real_engine.rs
```diff
@@ -437,8 +442,23 @@ impl<T: Transport, D: Dht, SecBuf: Buffer, Crypto: CryptoSystem> RealEngine<T, D
...
+                                    type_hint: String::new(),
+                                    aspect: vec![],
+                                    publish_ts: 0,
+                                };
+                                aspect_list.push(fake_aspect);
+                            }
+                            let fake_entry = EntryData {
+                                entry_address: entry_address.clone(),
+                                aspect_list,
+                            };
```
neonphog: why are they fake? we're not in test code here, are we?

crates/lib3h/src/engine/real_engine.rs
ddd-mtl: ❤️ thanks!

crates/lib3h/src/engine/real_engine.rs
ddd-mtl: That entryData is fake in the sense there is no actual content, but all the addresses are 'real'.
Didn't want to bother with adding a new struct for just addresses at this point.
What better term would you prefer?

neonphog: I get it. Ok, cool, yeah maybe just add a comment in here, and I'm fine with fake :+1:

ddd-mtl: 👍

Issue #141, opened 2019-07-02
First pass at converting current crypto system to new format


chrisgagne:
As a developer, I don't want to be distracted by verbose generics, so we'll use trait objects for the crypto system

Acceptance Criteria:

* We're using trait objects for the crypto system
* There are no generics in the crypto system
* Current CI passes

PR: https://github.com/holochain/lib3h/pull/155

PR #155, opened 2019-07-08
Implement CryptoSystem using Trait-Objects

neonphog:
## PR summary

RESOLVES https://github.com/holochain/lib3h/issues/141

This means we can replace:

```
impl<T: Transport, D: Dht, SecBuf: Buffer, Crypto: CryptoSystem> RealEngine<T, D, SecBuf, Crypto> {	
```
with
```
impl<T: Transport, D: Dht> RealEngine<T, D> {
```

## Review checklist 


crates/crypto_api/src/crypto_system/mod.rs
```diff
@@ -0,0 +1,77 @@
...
+    /// `public_key`
+    fn sign_verify(
+        &self,
+        signature: &Box<dyn Buffer>,
+        message: &Box<dyn Buffer>,
+        public_key: &Box<dyn Buffer>,
+    ) -> CryptoResult<bool>;
+}
+
+pub mod crypto_system_test;
```
ddd-mtl: I believe you can put this under `#[cfg(test)]` so it compiles only during `cargo test` .

neonphog: I tried that to start, but then the lib3h_sodium crate doesn't have access to it (even in tests)... If anyone knows a better way to organize this so the test code doesn't get build/exposed in the production lib, I'd love the feedback.

PR #156, opened 2019-07-08
Mock: DM edge cases

ddd-mtl:
## PR summary
Handle some SendMessage edge cases:
 - messaging self
 - messaging unknown agent

Refactored sendMessage code in real_engine.rs
Added test_send_message_fail()


crates/lib3h/src/engine/real_engine.rs
```diff
@@ -554,6 +503,58 @@ impl<T: Transport, D: Dht, SecBuf: Buffer, Crypto: CryptoSystem> RealEngine<T, D
...
+            request_id: msg.request_id.clone(),
+            space_address: msg.space_address.clone(),
+            to_agent_id: msg.from_agent_id.clone(),
+            result_info: vec![],
+        };
+        // Check if messaging self
+        let this_peer = space_gateway.this_peer();
+        if this_peer.peer_address.as_bytes() == msg.to_agent_id.as_slice() {
+            response.result_info = "Messaging self".as_bytes().to_vec();
+            return Lib3hServerProtocol::FailureResult(response);
```
struktured: Why is messaging yourself considered a failure?

crates/lib3h/tests/test_suites/two_basic.rs
```diff
@@ -100,7 +101,11 @@ fn test_send_message(alex: &mut NodeMock, billy: &mut NodeMock) {
     let req_id = alex.send_direct_message(&BILLY_AGENT_ID, "wah".as_bytes().to_vec());
     let (did_work, srv_msg_list) = alex.process().unwrap();
     assert!(did_work);
-    assert_eq!(srv_msg_list.len(), 0);
+    assert_eq!(srv_msg_list.len(), 1);
```
struktured: Why did the code previously return no messages? Was the send direct message code only half written?

crates/lib3h/tests/test_suites/two_basic.rs
```diff
@@ -126,6 +135,31 @@ fn test_send_message(alex: &mut NodeMock, billy: &mut NodeMock) {
...
+
+    // Send to unknown
+    let req_id = alex.send_direct_message(&CAMILLE_AGENT_ID, "wah".as_bytes().to_vec());
+    let (did_work, srv_msg_list) = alex.process().unwrap();
+    assert!(did_work);
+    assert_eq!(srv_msg_list.len(), 1);
+    println!("response: {:?}", srv_msg_list);
+    let msg_1 = &srv_msg_list[0];
+    one_let!(Lib3hServerProtocol::FailureResult(response) = msg_1 {
+        assert_eq!(response.request_id, req_id);
```
struktured: It might be worth distinguishing the unknown case somehow (an enum variant?) and verifying it in fact was unknown. This could have failed for some other unrelated reason, as far as the test is concerned.

crates/lib3h/src/engine/real_engine.rs
```diff
@@ -554,6 +503,58 @@ impl<T: Transport, D: Dht, SecBuf: Buffer, Crypto: CryptoSystem> RealEngine<T, D
...
+        let net_msg = if is_response {
+            P2pProtocol::DirectMessageResult(msg.clone())
+        } else {
+            P2pProtocol::DirectMessage(msg.clone())
+        };
+        // Serialize payload
+        let mut payload = Vec::new();
+        net_msg
+            .serialize(&mut Serializer::new(&mut payload))
+            .unwrap();
```
struktured: If this serialization errors happen easily I would suggest returning it back as a `FailureResult`. I don't know when that would actually occur though.

crates/lib3h/src/engine/real_engine.rs
ddd-mtl: Because the message will not be sent since you dont have a connection id to self, and there is no use-case for this. Allowing it would be a waste of cycles.

crates/lib3h/tests/test_suites/two_basic.rs
ddd-mtl: That part of the spec is unspecified. I'm pushing for every request from Core should get a response from Lib3h.

crates/lib3h/tests/test_suites/two_basic.rs
ddd-mtl: yeah, totally but our Error handling code isn't there yet...

crates/lib3h/src/engine/real_engine.rs
ddd-mtl: Agree. I believe serialization like this can't ever fail so no need for it. I believe it returns a Result because its a trait and that other implemetations could fail.

Issue #147, opened 2019-07-04
in-memory mock: Handle multiple spaces


ddd-mtl:
Should already work but we need to add tests to make sure handling of multiple spaces work properly in mock mode.

PR #157, opened 2019-07-08
Mock: Advanced Space handling + tests

ddd-mtl:
## PR summary
Solved #147 

Implemented handling of Lib3hClientProtocol::LeaveSpace.

Tests for leaving and rejoining a space.
Tests with multiple space.

Refactored some test code easy re-use in other tests.


crates/lib3h/src/engine/real_engine.rs
```diff
@@ -555,6 +554,28 @@ impl<T: Transport, D: Dht> RealEngine<T, D> {
...
+            to_agent_id: join_msg.agent_id.clone(),
+            result_info: vec![],
+        };
+        // Bail if space not joined by agent
+        let chain_id = (join_msg.space_address.clone(), join_msg.agent_id.clone());
+        if !self.space_gateway_map.contains_key(&chain_id) {
+            res.result_info = "Agent is not part of the space".to_string().into_bytes();
+            return Lib3hServerProtocol::FailureResult(res);
+        }
+        self.space_gateway_map.remove(&chain_id).unwrap();
```
struktured: Minor optimization but you could just call `remove` and check if the return value is `None`- if it is then produce the error result. Also, this code is simple enough that using a mutable reference for the result instance isn't a huge deal but it would be really easy to make this code immutable by building the result at the very end and using a conditional let statement to populate the `result_info` .

rustfmt.toml
```diff
@@ -28,7 +28,7 @@ combine_control_expr = true
 struct_field_align_threshold = 0
 match_arm_blocks = true
 force_multiline_blocks = false
-fn_args_density = "Tall"
+fn_args_layout = "Tall"
```
struktured: Looks innocuous but not sure why this change is in this PR.

ddd-mtl: Minor rust version update leftovers.

crates/lib3h/src/engine/real_engine.rs
ddd-mtl: @struktured Yeah, its a bit cleaner like you say even if the method is still not immutable.

PR #162, opened 2019-07-10
Mock: three nodes tests

### LINKED ISSUE #154: in-memory mock: three nodes tests
import three nodes tests from test_bin

ddd-mtl:
## PR summary
resolves #154 
Added TestSuite with for three nodes scenario.
Made some changes so nodes can gossip peer info from the network_gateway.
DHT does not automatically add peer received from gossip now. Must be confirmed by owner with HoldPeer command.
Since we have no connection management, real_engine will now try to connect to every peer it receives a HoldPeerRequested from the network_gateway.

### Misc
Added `process_count` for debugging/logging
Added global static for network_gateway identifer: `NETWORK_GATEWAY_ID`
Added more process() calls in the other tests since handshake has more back'n'forth now.

## Review checklist 


crates/lib3h/tests/utils/mod.rs
```diff
@@ -9,8 +9,15 @@ macro_rules! assert_process_success {
         assert!(did_work);
         assert_eq!(srv_msg_list.len(), 1);
         let msg_1 = &srv_msg_list[0];
-        one_let!(Lib3hServerProtocol::SuccessResult(response) = msg_1 {
+        if let Lib3hServerProtocol::SuccessResult(response) = msg_1 {
             assert_eq!(response.request_id, $req);
-        });
+        } else {
+            if let Lib3hServerProtocol::FailureResult(response) = msg_1 {
```
neonphog: since there's multiple `if let`s, should this be a `match` insteadL

ddd-mtl: good point. I'll change that.

PR #165, opened 2019-07-11
Makefile env tweak

neonphog:
## PR summary

Tiny change. Now, instead of running `make -f Makefile.local`, I can just put `MAKE_ENV="local"` in my .bashrc and run `make`. Default without that env var is still to run the nix versions of tests / commands.

## Review checklist 

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,12 @@
-.PHONY: all test fmt clean
+# Top-Level Environment Switcher
+# This will delegate to sub Makefile.* files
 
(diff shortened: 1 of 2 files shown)


neonphog:
@ddd-mtl 

> ok.
> I would probably prefer local as the default and have MAKE_ENV="nix", but I'm not using make for now so can't really tell from a user perspective.

I agree personally, but for consistency across the org, we're being asked to have nix be the default.

PR #181, opened 2019-07-11
busywork fleshing out crypto-system apis: hashing functions

neonphog:
## Review checklist 

### DIFF
--- a/crates/crypto_api/src/crypto_system/crypto_system_test.rs
+++ b/crates/crypto_api/src/crypto_system/crypto_system_test.rs
@@ -15,6 +15,7 @@ impl FullSuite {
     pub fn run(&self) {
         self.test_sec_buf();
         self.test_random();
+        self.test_pwhash();
         self.test_sign_keypair_sizes();
         self.test_sign_keypair_generation();
         self.test_sign();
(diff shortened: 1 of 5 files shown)


crates/crypto_api/src/crypto_system/crypto_system_test.rs
```diff
@@ -251,6 +278,74 @@ mod test {
...
+            hash: &mut Box<dyn Buffer>,
+            data: &Box<dyn Buffer>,
+        ) -> CryptoResult<()> {
+            if hash.len() != self.hash_sha256_bytes() {
+                return Err(CryptoError::BadHashSize);
+            }
+
+            let mut hasher = sha2::Sha256::new();
+            hasher.input(data.read_lock().deref());
+            hash.write(0, &hasher.result())?;
```
struktured: Just want to make sure you made a conscious decision to unwrap here as opposed to returning the error back in the `CryptoResult`. Same with the other implemented hash functions. I don't know why the write here would even fail- perhaps you already checked for it by verifying the hash lengths are correct?

neonphog: @struktured - I don't see an unwrap here. The `?` will forward any write errors up to the caller.

PR #184, opened 2019-07-12
cleanup repo root by moving build files into build directory

neonphog:
## PR summary

There's no substantive change here, just putting build related files together in a `build` directory to keep our repo root clean.

## Review checklist 


neonphog:
@struktured - good point, cmake best practice is to create a `build` directory and cd into it for building.

How about `scripts`? I think that also gives a better indication of what's inside it, and I've seen it as a precedent in other opensource repos.

ddd-mtl:
@neonphog @neonphog how about `/build_scripts`? :)

PR #185, opened 2019-07-12
Mock: fixed rust tech debt

### LINKED ISSUE #179: Solve Rust puzzle concerning post_dht()
Use `Dht::post()` instead of `.post_dht()` hack

ddd-mtl:
## PR summary

Resolves #179 

Moved `P2pGateway::post_dht()` code to `Dht::post()` implementation for P2pGateway.

It seems the newer rust version fixed the problem?

### DIFF
--- a/crates/lib3h/src/gateway/gateway_dht.rs
+++ b/crates/lib3h/src/gateway/gateway_dht.rs
@@ -2,7 +2,7 @@
 
 use crate::{
     dht::{dht_protocol::*, dht_trait::Dht},
-    engine::p2p_protocol::*,
+    engine::{p2p_protocol::*, NETWORK_GATEWAY_ID},
     gateway::{self, P2pGateway},
     transport::transport_trait::Transport,
 };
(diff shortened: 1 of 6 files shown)


crates/lib3h/src/gateway/gateway_dht.rs
```diff
@@ -32,7 +32,29 @@ impl<T: Transport, D: Dht> Dht for P2pGateway<T, D> {
 
     /// Processing
     fn post(&mut self, cmd: DhtCommand) -> Lib3hResult<()> {
-        // TODO #179
+        // Add to connection_map for space_gateways
+        // TODO #176 - Maybe we shouldn't have different code paths for populating
+        // the connection_map between space and network gateways.
+        if self.identifier != NETWORK_GATEWAY_ID {
```
struktured: A little confused about why the network gateway if is constant and how the if statement determines that it's a space gateway. Can you explain this a bit more?

ddd-mtl: P2pGateway code is used for both space and network gateways.
In the RealEngine, there is only one network gateway but potentially many space gateways.
So in P2pGateway code, if the gateway identifier is not the unique network gateway identifier, than its some space gateway.
Ideally we don't want different code behavior between space and network gateways, thats the whole point of this P2pGateway struct, but as of now I do have different behaviors because the connection is handled differently and needs to be sorted out more when implementing the disconnection/reconnection.
Hope thats clear.

PR #186, opened 2019-07-12
Custom error implmentation / Moving away from Failure crate

dymayday:
## PR summary

This PR is related to #58 and is still a work in progress.

## Review checklist 


crates/lib3h_protocol/src/error.rs
```diff
@@ -0,0 +1,105 @@
...
+pub enum ErrorKind {
+    /// An I/O error that occurred while processing a data stream.
+    Io(io::Error),
+    /// Error occuring when using `transport`.
+    TransportError(String),
+    /// An error occuring whiling trying to deserialize stuff during gossiping for example.
+    // DeserializeError(String),
+    DeserializeError(DeserializeError),
+    /// Yet undefined error.
+    Other(String),
```
neonphog: Is this common naming? In our org, I've seen this called a couple things, including `Generic(String),`. If there's a rust-world standard naming, I'm cool with it, or if not, we should try to pick something and stick with it.

neonphog:
This is looking awesome as a start @dymayday ! I love the cleanup of the error types.

I don't see an easy way to convert a Lib3Result/Error into a Lib3ProtocolResult/Error. Which may be totally ok, it's hard for me to tell without seeing the holochain-rust integration side of this.

The thing I'd like to make sure is: for the most part, core crates only need to pull in lib3h_protocol. The only crates that need lib3h are conductor, net, test_bin etc crates that are dealing directly with RealEngine and other lib3h structs need to work with the Lib3hResult/Error.

Nice work!

crates/lib3h/src/error.rs
```diff
@@ -0,0 +1,125 @@
...
+    /// An I/O error that occurred while processing a data stream.
+    Io(io::Error),
+    /// Error occuring when using `transport`.
+    TransportError(TransportError),
+    /// Error originating from [lib3h_protocol] crate.
+    Lib3hProtocolError(Lib3hProtocolError),
+    /// Error occuring from [Hcid](hcid) crate.
+    HcId(hcid::HcidError),
+    /// Error originating from [MessagePack](rmp_serde) deserializing crate.
+    RMPSerdeDecodeError(RMPSerdeDecodeError),
```
ddd-mtl: Can you change case to `RmpSerdeDecodeError` please? (here and all the other places with RMP)

crates/lib3h/src/error.rs
```diff
@@ -0,0 +1,125 @@
...
+    /// Error from the [lib3h_crypto_api] crate.
+    CryptoAPIError(CryptoError),
```
ddd-mtl: Can you change case to `CryptoApiError` please?

crates/lib3h/src/engine/network_layer.rs
```diff
@@ -188,7 +190,11 @@ impl<T: Transport, D: Dht> RealEngine<T, D> {
                 let space_gateway = self
                     .space_gateway_map
                     .get_mut(&(msg.space_address.to_owned(), msg.to_peer_address.to_owned()))
-                    .ok_or_else(|| format_err!("space_gateway not found"))?;
+                    .ok_or_else(|| {
+                        Lib3hError::new(ErrorKind::KeyNotFound(String::from(
+                            "space_gateway not found",
+                        )))
+                    })?;
```
ddd-mtl: This seems way to verbose for me.
Can we have a follow-up PR with converters or a macro that would allow us to write something link:
`lib3h_err!(KeyNotFound, ""space_gateway not found")` ?

crates/lib3h/src/error.rs
```diff
@@ -0,0 +1,125 @@
...
+
+/// The specific type of an error.
+#[derive(Debug)]
+pub enum ErrorKind {
+    /// An I/O error that occurred while processing a data stream.
+    Io(io::Error),
+    /// Error occuring when using `transport`.
+    TransportError(TransportError),
+    /// Error originating from [lib3h_protocol] crate.
+    Lib3hProtocolError(Lib3hProtocolError),
```
ddd-mtl: I believe you got this backwards.
`Lib3hProtocolError` will most probably not have its own errors since its just an api (declarations only).
So I see `Lib3hError` as an ErrorKind of `Lib3hProtocolError` and not the other way around.

ddd-mtl:
@neonphog I believe we should aim for only `net` crate using `lib3h`. `test_bin` is going to be deprecated anyway. I don't see why `conductor` would use it. Maybe we should move `RealEngineConfig` to `lib3h_protocol` and rename it `NetworkEngineConfig` ?

crates/lib3h/src/error.rs
dymayday: I understand and this is what I tried first but it gives us circular dependency trouble, as `Lib3h` already depends from `Lib3hProtocol` so I cannot make `Lib3hProtocol` depends from `Lib3h` as well :/

This gave me a little trouble and this is the solution I found. But I'm of course open to discuss it further :)

crates/lib3h/src/error.rs
dymayday: :+1:

crates/lib3h/src/error.rs
dymayday: :+1:

crates/lib3h_protocol/src/error.rs
dymayday: The standard naming way is indeed to use [Other](https://doc.rust-lang.org/std/io/enum.ErrorKind.html#variant.Other).
But I'm not too picky about this either :)

crates/lib3h/src/engine/network_layer.rs
dymayday: Yeah I agree, this is kinda too verbose and that's on my list of improvement :)

crates/lib3h_protocol/src/error.rs
ddd-mtl: I'm fine with `Other`.

PR #188, opened 2019-07-12
Mock: cleaner api

ddd-mtl:
## PR summary
Removed run/stop/terminate() from NetworkEngine trait.
Moved terminate() to a non-trait shutdown() function called on drop.
Renamed some TransportEvents for better clarity of what happened.
Fixed bug of Wss Engine not using NETWORK_GATEWAY_ID.


crates/lib3h/src/engine/real_engine.rs
```diff
@@ -187,8 +174,25 @@ impl<T: Transport, D: Dht> NetworkEngine for RealEngine<T, D> {
     }
 }
 
+/// Drop
```
struktured: Do we need tests for the `Drop` traits or is this tested naturally by virtue of any transport test successfully finishing?

ddd-mtl: yes, exactly. You'd see that warn log in the tests logs if something errored.

PR #190, opened 2019-07-13
Mock disconnect: IncomingConnectionEstablished & ConnectionClosed for TransportMemory

ddd-mtl:
## PR summary
Differentiated `IncomingConnectionEstablished` from `ConnectResult`
Refactored handling of `ConnectResult` in its own function so match on `IncomingConnectionEstablished` has the same behavior.
RealEngine's shutdown now closes all connections an all gateways.
Changed MemoryServer to receive closing connections events.
Clarified a bit MemoryServer and TransportMemory code with better comments and some field renaming.


crates/lib3h/src/engine/network_layer.rs
```diff
@@ -85,6 +85,57 @@ impl<T: Transport, D: Dht> RealEngine<T, D> {
...
+                let data = ConnectedData {
+                    request_id: "FIXME".to_string(), // TODO #173
+                    uri,
+                };
+                outbox.push(Lib3hServerProtocol::Connected(data));
+            }
+            let _ = self.network_connections.insert(id.to_owned());
+        }
+        Ok(outbox)
+    }
```
ddd-mtl: no new code here, just moved it out of the match

crates/lib3h/src/gateway/gateway_transport.rs
```diff
@@ -180,6 +180,47 @@ impl<T: Transport, D: Dht> P2pGateway<T, D> {
...
+            .serialize(&mut Serializer::new(&mut buf))
+            .unwrap();
+        trace!(
+            "({}) sending P2pProtocol::PeerAddress: {:?} to {:?}",
+            self.identifier.clone(),
+            our_peer_address,
+            id
+        );
+        return self.inner_transport.borrow_mut().send(&[&id], &buf);
+    }
```
ddd-mtl: same here

crates/lib3h/src/transport/memory_mock/memory_server.rs
```diff
@@ -94,28 +97,32 @@ impl MemoryServer {
         if res.is_some() {
             return Err(TransportError::new("connectionId already used".to_string()));
         }
-        // Notify our TransportMemory to connect back
-        self.new_conn_inbox.push(requester_uri.to_string());
-        self.connection_ids.insert(requester_uri.to_string());
+        // Notify our TransportMemory (so it can connect back)
+        self.connection_inbox
+            .push((requester_uri.to_string(), true));
```
neonphog: could convert from tuple to struct for clarity on what the bool means, but not a big deal

ddd-mtl: how about just typedefs for true and false?

PR #191, opened 2019-07-13
Transport wss test suite

struktured:
## PR summary

This PR leverages the existing real engine test suite for the websocket transports (both tls and unencrypted).

- [x] unencrypted websocket test suite
- [x] tls encrypted websocket test suite

## Review checklist 


struktured:
> sweet!
> In the future, we might make a macro out of the `launch_*_test()` since its always the same thing, don't you think?

Agreed- there is still some boiler plate left here to eliminate but it's a good start!

PR #192, opened 2019-07-13
busywork fleshing out crypto-system apis: key derivation, key exchange, and aead

neonphog:
## PR summary

- finishes off crypto system api!
- also marks old api deprecated

## Review checklist 


crates/crypto_api/src/crypto_system/mod.rs
```diff
@@ -48,11 +54,23 @@ pub trait CryptoSystem: Sync {
...
+    fn kdf_max_bytes(&self) -> usize;
+
+    /// derive a new deterministic key based of index, context, and parent
+    fn kdf(
+        &self,
+        out_buffer: &mut Box<dyn Buffer>,
+        index: u64,
+        context: &Box<dyn Buffer>,
+        parent: &Box<dyn Buffer>,
+    ) -> CryptoResult<()>;
```
struktured: It seems a little too easy to use the wrong buffer here in the argument list, maybe in a future iteration we switch to a builder pattern for this?

crates/crypto_api/src/crypto_system/mod.rs
```diff
@@ -100,6 +118,85 @@ pub trait CryptoSystem: Sync {
...
+
+    fn aead_encrypt(
+        &self,
+        cipher: &mut Box<dyn Buffer>,
+        message: &Box<dyn Buffer>,
+        adata: Option<&Box<dyn Buffer>>,
+        nonce: &Box<dyn Buffer>,
+        secret: &Box<dyn Buffer>,
+    ) -> CryptoResult<()>;
+
```
struktured: Same as above comment- builder pattern or something (wrapper types? a struct with fields equal to the existing function args?) to make this less error prone

crates/crypto_api/src/crypto_system/crypto_system_test.rs
```diff
@@ -446,5 +656,222 @@ mod test {
...
+                return Err(CryptoError::BadNonceSize);
+            }
+
+            if secret.len() != self.aead_secret_bytes() {
+                return Err(CryptoError::BadSecretKeySize);
+            }
+
+            let cipher = cipher.read_lock();
+
+            if &cipher[2..4] != &nonce.read_lock()[..2] {
```
struktured: This code is cryptic, maybe deserves a comment about what is going here with the seemingly arbitrary array indexing.

crates/sodium/src/hash.rs
```diff
@@ -14,6 +14,7 @@ pub const BYTES512: usize = rust_sodium_sys::crypto_hash_sha512_BYTES as usize;
 /// @param {SecBuf} input - the data to hash
 ///
 /// @param {SecBuf} output - Empty Buffer to be used as output
+#[deprecated(note = "use CryptoSystem")]
```
struktured: What is purpose of this? To avoid ever breaking hc core? Are you going to audit hc core for deprecation warnings instead of build errors and make a separate PR in that repo?

neonphog: exactly, core uses a bunch of these legacy apis still, it's not a trivial amount of work to convert them over to crypto system..

crates/crypto_api/src/crypto_system/crypto_system_test.rs
neonphog: :+1:

crates/crypto_api/src/crypto_system/mod.rs
neonphog: named param macros!

crates/crypto_api/src/crypto_system/mod.rs
neonphog: named param macros!

PR #196, opened 2019-07-15
bring makefile rust versions inline with nix

neonphog:
## PR summary

@ddd-mtl FYI - this bumps makefile rust version to `nightly-2019-07-14` bringing it inline with nix here - https://github.com/holochain/lib3h/pull/193

## Review checklist 


neonphog:
In this particular case it was failing because the appveyor.yml wasn't in this branch yet. Also, we seem to be getting stuck behind holochain-rust builds in the queue... I think we should not mark it required for merge until we see how stable it is && if we can get a separate queue for builds

PR #201, opened 2019-07-15
Mock: timestamp of peerData

ddd-mtl:
## PR summary
Added `since_epoch_ms()` function in new `time.rs` file.
PeerData of this is created with proper timestamp, than gossiped around.

### DIFF
--- a/crates/lib3h/src/engine/network_layer.rs
+++ b/crates/lib3h/src/engine/network_layer.rs
@@ -215,13 +215,14 @@ impl<T: Transport, D: Dht> RealEngine<T, D> {
                     Dht::post(&mut *self.network_gateway.borrow_mut(), cmd)?;
                 } else {
                     // otherwise should be for one of our space
-                    let space_gateway = self
+                    let maybe_space_gateway = self
                         .space_gateway_map
-                        .get_mut(&(msg.space_address.to_owned(), msg.to_peer_address.to_owned()))
-                        .ok_or_else(|| {
-                            format_err!("space_gateway not found: {}", msg.space_address)
-                        })?;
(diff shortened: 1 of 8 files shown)


crates/lib3h/src/engine/network_layer.rs
```diff
@@ -215,13 +215,17 @@ impl<T: Transport, D: Dht> RealEngine<T, D> {
...
-                    Dht::post(space_gateway, cmd)?;
+                        .get_mut(&(msg.space_address.to_owned(), msg.to_peer_address.to_owned()));
+                    if let Some(space_gateway) = maybe_space_gateway {
+                        Dht::post(space_gateway, cmd)?;
+                    } else {
+                        warn!(
+                            "received gossip for unjoined space_gateway: {}",
+                            msg.space_address
+                        );
+                    }
```
ddd-mtl: We actually can receive gossip of unjoined space because other nodes are not notified when we unjoin.
The timestamp triggers this now because now when rejoining it sees an updated timestamp and so tries to gossip stuff back to the "new" peer.

crates/lib3h/src/gateway/gateway_transport.rs
```diff
@@ -277,7 +281,7 @@ impl<T: Transport, D: Dht> P2pGateway<T, D> {
                             let peer = PeerData {
                                 peer_address: peer_address.clone(),
                                 peer_uri: transport_id_to_url(id.clone()),
-                                timestamp: 42, // TODO #166
+                                timestamp: peer_timestamp,
```
neonphog: I think this works for now... eventually we'll probably need to separate the timestamps for the spaces from those for the transport layer... but we can design that when we start getting into rrdht

PR #202, opened 2019-07-16
removing the update

duffybelfield:
## PR summary

Minimising the AV config

### DIFF
--- a/scripts/windows/do-ci-test.bat
+++ b/scripts/windows/do-ci-test.bat
@@ -2,8 +2,5 @@
 setlocal enabledelayedexpansion
 
 rem KEEP IN SYNC WITH HOLONIX
-set nightly-date=nightly-2019-07-14
-rustup toolchain install --no-self-update !nightly-date!
-rustup default !nightly-date!
 
 cargo test --verbose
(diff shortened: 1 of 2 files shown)


scripts/windows/do-ci-test.bat
```diff
@@ -2,8 +2,4 @@
 setlocal enabledelayedexpansion
 
 rem KEEP IN SYNC WITH HOLONIX
-set nightly-date=nightly-2019-07-14
-rustup toolchain install --no-self-update !nightly-date!
-rustup default !nightly-date!
-
-cargo test --verbose
+cargo test --release -p lib3h --target-dir c:\build\lib3h\target --verbose
```
neonphog: @duffybelfield - couple things:

1) There is a subcrate here called `lib3h`... passing `-p lib3h` will cause only that crate to be tested... we need to leave that off so all crates are tested.

2) Without the rustup installs, there might be a period after we update the nightly version, but before the lib3h-hot image is updated where builds will fail because of version mismatch. Thoughts?

neonphog: 3) I'd prefer to leave the --release off, so if there's errors we get relevant stack information... actually `RUST_BACKTRACE=1 cargo test --verbose` might be better.

neonphog: ^^ not positive how to set environment variables in windows lol : )

PR #210, opened 2019-07-16
Mock: disconnect / reconnect  & PeerTimedOut

ddd-mtl:
### Disconnection/ Reconnection
Reworked MemoryServer and TransportMemory to handle disconnections and reconnections properly.
#### MemoryServer
Clarified code to distinguish connection directions (inbound, outbound). All connectionIds are actually inbound.
#### TransportMemory
(Moved Drop)
Rewrote `process()` as its important to process `IncomingConnectionEstablished` first before anything else coming from our own MemoryServer.

### Peer TimedOut

Added `gossip_interval` and `timeout_threshold` settings to the DHT config.
The mirrorDht will gossip `this_peer` at each gossip interval.
On each `process()` check all peers if they timed out according to `timeout_threshold`.
Send a `DhtEvent::PeerTimedOut` every time it occurs.
mirrorDht being monotonic **even for peerData** we keep track of timed-out peers so that we don't spam `DhtEvent::PeerTimedOut` on every process.

### NodeMock

Added `disconnect()` and `reconnect()` methods.
`disconnect()` actually recreates the same engine, but this will call the drop that closes all connections.
It has an extra step of creating a temporary dummy Engine, to make sure the MemoryServer is unset.
 `reconnect()` tries to connect to all previously URIs given in `connect()`, and also rejoins all known spaces.

### Misc
Added `get_other_peer_list()` for DRYness.

Added `two_connection` test suite.

Fixed Dht unit test that had hardcoded timestamps.


crates/lib3h/src/gateway/gateway_dht.rs
```diff
@@ -102,28 +104,13 @@ impl<T: Transport, D: Dht> P2pGateway<T, D> {
                     p2p_gossip
                         .serialize(&mut Serializer::new(&mut payload))
                         .unwrap();
```
dymayday: Do you think that we can have an `expect` here instead of a `unwrap`? (it's most of the time easier to debug later IMHO :) )

crates/lib3h/src/gateway/p2p_gateway.rs
```diff
@@ -42,6 +42,36 @@ impl<T: Transport, D: Dht> P2pGateway<T, D> {
...
+        );
+        // get connection_id
+        let maybe_connection_id = self.connection_map.get(&peer_uri);
+        if maybe_connection_id.is_none() {
+            return None;
+        }
+        let conn_id = maybe_connection_id.unwrap().clone();
+        trace!(
+            "({}) get_connection_id: {} -> {} -> {}",
+            self.identifier.clone(),
```
dymayday: Can we use a `&self.identifier` here instead ?

crates/lib3h/src/time.rs
```diff
@@ -1,25 +1,36 @@
...
 
+pub static mut START_TIME: SystemTime = UNIX_EPOCH;
+
 pub fn since_epoch_ms() -> u64 {
-    let start = SystemTime::now();
-    let since_the_epoch = start
-        .duration_since(UNIX_EPOCH)
-        .expect("Time went backwards");
+    let since_the_epoch;
+    unsafe {
```
dymayday: Do you happen to know why we need this in an `unsafe` block ?

crates/lib3h/src/transport/memory_mock/transport_memory.rs
```diff
@@ -115,59 +137,75 @@ impl Transport for TransportMemory {
...
-            return Err(TransportError::new(format!(
-                "No known connection for connectionId {}",
-                id
-            )));
+        let maybe_other_uri = self.outbound_connection_map.get(id);
+        if let None = maybe_other_uri {
+            return Err(TransportError::new(format!("Unknown connectionId: {}", id)));
         }
-        let url = maybe_url.unwrap();
+        let other_uri = maybe_other_uri.unwrap();
```
dymayday: Do you know if we can use an `expect`, a `?` or a default value here instead of an `unwrap` ?

crates/lib3h/src/transport/memory_mock/transport_memory.rs
```diff
@@ -115,59 +137,75 @@ impl Transport for TransportMemory {
...
             return Err(TransportError::new(format!(
                 "No Memory server at this url: {}",
-                url,
+                other_uri,
             )));
         }
-        let mut server = maybe_server.unwrap().lock().unwrap();
-        // Tell it to close connection
-        server.close(&id)?;
+        let mut other_server = maybe_other_server.unwrap().lock().unwrap();
```
dymayday: Using something like a
```rust
let mut other_server = maybe_other_server.unwrap_or({ ... }).lock().expect("Fail to lock...");
```
could be easier to read, what do you think?

crates/lib3h/src/transport/memory_mock/transport_memory.rs
```diff
@@ -115,59 +137,75 @@ impl Transport for TransportMemory {
...
+        if self.maybe_my_uri.is_none() {
+            return Err(TransportError::new(
+                "Cannot send before bounding".to_string(),
+            ));
+        }
+        let my_uri = self.maybe_my_uri.clone().unwrap();
         for id in id_list {
             // Get the other node's uri on that connection
-            let maybe_uri = self.connection_map.get(*id);
+            let maybe_uri = self.outbound_connection_map.get(*id);
```
dymayday: What do you think about: 
```rust
let maybe_uri = self.outbound_connection_map.get(*id).unwrap_or({
    warn!("No known connection for connectionId: {}", id);
    continue
}
```

crates/lib3h/src/transport/memory_mock/transport_memory.rs
```diff
@@ -115,59 +137,75 @@ impl Transport for TransportMemory {
...
+            if let Err(e) = res {
+                warn!("Closing connection {} failed: {:?}", id, e);
+            }
         }
         Ok(())
     }
 
     /// Send payload to known connectionIds in `id_list`
     fn send(&mut self, id_list: &[&ConnectionIdRef], payload: &[u8]) -> TransportResult<()> {
+        if self.maybe_my_uri.is_none() {
```
dymayday: Same thing here with `unwrap_or`, what do you think?

crates/lib3h/src/transport/memory_mock/transport_memory.rs
```diff
@@ -233,53 +271,75 @@ impl Transport for TransportMemory {
...
+        let mut output = Vec::new();
+        for my_server_uri in &self.my_servers {
             let server_map = memory_server::MEMORY_SERVER_MAP.read().unwrap();
-            let server = server_map.get(server_uri).expect("My server should exist.");
-            let (success, output) = server.lock().unwrap().process()?;
+            let mut my_server = server_map
+                .get(my_server_uri)
+                .expect("My server should exist.")
+                .lock()
+                .unwrap();
```
dymayday: Do you think an `expect` could be useful here, or it's meaningless after a lock?

crates/lib3h/src/transport/memory_mock/transport_memory.rs
```diff
@@ -233,53 +271,75 @@ impl Transport for TransportMemory {
...
+                    let out_cid = self
+                        .inbound_connection_map
+                        .get(&in_cid)
+                        .expect("Should have outbound at this stage");
+                    outbox.push(TransportEvent::ReceivedData(
+                        out_cid.to_string(),
+                        data.clone(),
+                    ));
+                }
+                _ => unreachable!(),
```
dymayday: It's a matter of preference, but by removing this `unreachable!()` statement, and so removing the default match pattern, we can catch some missing implementation of possible future additional enum variant. What do you think?

ddd-mtl: There is no possiblity of missing future additional enum variant because it would have to be implemented by the MemoryServer first. Here we are only reacting to whatever the MemoryServer can send us. Will add a comment to clarify.

crates/lib3h/src/dht/mirror_dht.rs
```diff
@@ -238,10 +328,15 @@ impl MirrorDht {
...
-                    MirrorGossip::Peer(peer) => {
-                        let maybe_peer = self.get_peer(&peer.peer_address);
-                        if maybe_peer.is_none() {
-                            return Ok(vec![DhtEvent::HoldPeerRequested(peer)]);
+                    MirrorGossip::Peer(gossiped_peer) => {
+                        let maybe_known_peer = self.get_peer(&gossiped_peer.peer_address);
+                        if maybe_known_peer.is_none() {
+                            return Ok(vec![DhtEvent::HoldPeerRequested(gossiped_peer)]);
+                        }
+                        let known_peer = maybe_known_peer.unwrap();
```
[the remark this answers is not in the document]
ddd-mtl: Rust doesnt allow what you are suggesting:
```
warning: unreachable expression
   --> crates\lib3h\src\dht\mirror_dht.rs:337:42
    |
337 |                           let known_peer = maybe_known_peer.unwrap_or(
    |  __________________________________________^
338 | |                             return Ok(vec![DhtEvent::HoldPeerRequested(gossiped_peer)])
339 | |                         );
    | |_________________________^
    |
    = note: #[warn(unreachable_code)] on by default

warning: unused variable: `known_peer`
   --> crates\lib3h\src\dht\mirror_dht.rs:337:29
    |
337 |                         let known_peer = maybe_known_peer.unwrap_or(
    |                             ^^^^^^^^^^ help: consider prefixing with an underscore: `_known_peer`
    |
    = note: #[warn(unused_variables)] on by default

    Finished dev [unoptimized + debuginfo] target(s) in 6.00s
     Running C:\github\lib3h-mock\target\debug\deps\lib3h-93067470e9c0c5ab.exe
```

crates/lib3h/src/transport/memory_mock/transport_memory.rs
ddd-mtl: doesn't work in Rust like previous comment.

crates/lib3h/src/transport/memory_mock/transport_memory.rs
ddd-mtl: doesn't work in Rust like previous comment.

crates/lib3h/src/transport/memory_mock/transport_memory.rs
ddd-mtl: doesn't work in Rust like previous comment.

crates/lib3h/src/time.rs
ddd-mtl: Because using a global variable is unsafe and START_TIME is one.

crates/lib3h/src/transport/memory_mock/transport_memory.rs
ddd-mtl: no because we want to return an error. Rust is missing a `unwrap_or_return()`

ddd-mtl: No because we want to return an error but there is no `map_err` for an Option. Rust is also missing an `unwrap_or_return()`

crates/lib3h/src/dht/mirror_dht.rs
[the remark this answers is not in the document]
dymayday: Sorry, I got confused, what about something like:
```rust

                    MirrorGossip::Peer(gossiped_peer) => {
                        if let Some(known_peer) = self.get_peer(&gossiped_peer.peer_address) {
                            // Update Peer timestamp
                            if gossiped_peer.timestamp > known_peer.timestamp {
                                let _ = self.add_peer(&gossiped_peer);
                            }
                            return Ok(vec![]);
                        } else {
                            return Ok(vec![DhtEvent::HoldPeerRequested(gossiped_peer)]);
                        }
                    }
```

crates/lib3h/src/transport/memory_mock/transport_memory.rs
dymayday: I meant at compile time, the Rust compiler can tell you that there is a missing branch. It's useful during refactoring. In the end it's just a nice to have, no biggie :)

crates/lib3h/src/time.rs
dymayday: Oh you mean a mutable global static one, this is indeed really unsafe ^^'

dymayday: Humm I'm not sure about what you are trying to do then ^^' 

Could you please tell me more about what this ```rust pub fn since_epoch_ms() ``` function is supposed to do ?

crates/lib3h/src/dht/mirror_dht.rs
```diff
@@ -29,32 +31,44 @@ pub struct MirrorDht {
     inbox: VecDeque<DhtCommand>,
     /// Storage of EntryData with empty aspect content?
     entry_list: HashMap<Address, HashSet<Address>>,
-    /// Storage of PeerData
-    peer_list: HashMap<String, PeerData>,
+    /// Monotonic Storage of PeerData
+    peer_map: HashMap<String, PeerData>,
+    /// Track if peer timed out
+    timed_out_map: HashMap<String, HasTimedOut>,
```
struktured: You could use a `HashSet<String>` instead for simplicity here, but if you intend to store more info in the future this pattern might be preferred.

crates/lib3h/src/dht/mirror_dht.rs
```diff
@@ -29,32 +31,44 @@ pub struct MirrorDht {
...
 }
 
 /// Constructors
 impl MirrorDht {
-    pub fn new(peer_address: &str, peer_transport: &Url) -> Self {
-        MirrorDht {
+    pub fn new(peer_address: &str, peer_uri: &Url) -> Self {
+        let dht_config = DhtConfig::new(peer_address, peer_uri);
+        Self::new_with_config(&dht_config).expect("Failed creating default MirrorDht")
+    }
```
struktured: Maybe just return `NetResult<Self>` self, here?

crates/lib3h/src/dht/mirror_dht.rs
```diff
@@ -119,21 +135,94 @@ impl Dht for MirrorDht {
...
+        for (peer_address, peer) in self.peer_map.iter() {
+            // Skip self
+            if peer_address == &self.this_peer.peer_address {
+                continue;
+            }
+            // Skip already timed out
+            let has_timed_out = self
+                .timed_out_map
+                .get(peer_address)
+                .expect("Should allways have time_out value for a peer");
```
struktured: Typo: `allways` -> `always`.

crates/lib3h/src/dht/mirror_dht.rs
```diff
@@ -119,21 +135,94 @@ impl Dht for MirrorDht {
...
+                continue;
+            }
+            // Check if timed-out
+            if now - peer.timestamp > self.config.timeout_threshold {
+                debug!("@MirrorDht@ peer {} timed-out", peer_address);
+                outbox.push(DhtEvent::PeerTimedOut(peer_address.clone()));
+                timed_out_list.push(peer_address.clone());
+                did_work = true;
+            }
+        }
```
struktured: Walking the peer map is a bit expensive. You could use a BTreeMap<u64, String> and invoke the [range](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.range) function over the time interval of interest. You just need to make sure to update the map every time a node communicates.

Another way I might have optimized this in a past life is by maintaining a low priority background thread which periodically does these checks outside of the critical process loop but I rather not introduce concurrency to our framework yet.

ddd-mtl: yeah, and not worth the effort for mirrorDHT as its temp stuff until we get rrdht.

crates/lib3h/src/dht/mirror_dht.rs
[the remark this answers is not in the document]
ddd-mtl: I find your proposal harder to read, as it has more nesting and the error case is done last.

PR #216, opened 2019-07-19
Clarify code

ddd-mtl:
## PR summary

Clarified / simplified some parts.
Added a peerAddress type

### DIFF
--- a/crates/lib3h/src/dht/dht_protocol.rs
+++ b/crates/lib3h/src/dht/dht_protocol.rs
@@ -1,6 +1,6 @@
+use crate::dht::PeerAddress;
 use lib3h_protocol::{data_types::EntryData, Address};
 use url::Url;
-use crate::dht::PeerAddress;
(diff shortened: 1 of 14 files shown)


crates/lib3h/src/dht/dht_protocol.rs
```diff
@@ -1,7 +1,8 @@
+use crate::dht::PeerAddress;
 use lib3h_protocol::{data_types::EntryData, Address};
 use url::Url;
 
-pub type FromPeerAddress = String;
+pub type FromPeerAddress = PeerAddress;
```
struktured: Not sure how useful `FromPeerAddress` is at all. I think we should use a struct with a field of that name instead. Not related directly to this PR however.

ddd-mtl: Yeah, I'm also on the fence for this one now

PR #221, opened 2019-07-25
RFC: add unique engine ids to logs

neonphog:
## PR summary

Let me know which parts if any of this you think would be useful...

- it's hard to tell which RealEngine instance is outputting logs, they all have the same name
- adds an id that is incremented in a lazy_static
- adds a helper macro, so we don't have to pass in / format that every time

If we want to go further with this, we can pass the d_id down into the submodule structs when they are created, and hopefully just re-use the same log macros... I didn't want to go further until/unless we agree this was worth it.

```
[DEBUG] <[1]> mem-agent handle_spaceDhtEvent...
[DEBUG] <[2]> mem-agent handle_spaceDhtEvent...
```

also, the `<[1]>` is green, and the `<[2]>` is yellow.

## Review checklist 


ddd-mtl:
The name field was made exactly for this purpose.
I believe the real problem here is that they all have the same name. They are not suppose to, and should have unique names.
Maybe change the type of 'name' to that DId newtype and have it auto-assign and increment if no name is provided in constructor config.
Also, I suggest renaming the DId type to DebugId. It will help us not forget to strip out this field when compiling for production.

neonphog:
Closing, we can revisit after core logging story settles

PR #234, opened 2019-08-05
Unique bind url per mem transport

struktured:
## PR summary

This modifies the `bind` function of the memory transport such that each transport instance will decorate the given `url` to bind with its own unique id. This fixes several logical problems otherwise introduced without it when connecting the same bind address multiple times. It also aids with debugging by distinguishing transport addresses easier.

## Review checklist 


struktured:
Reconnect/disconnect is broken but I'll fix post merge.

Issue #173, opened 2019-07-11
Track Connect requests


ddd-mtl:
As lib3h client, I can receive a request_id back (in the ConnectResult) from my own connect request so that I know I'm connected to another lib3h client as a result of that specific previous request.

Tasks:
- [ ] Keep track of `Lib3hClientProtocol::Connect.request_id` received so we can associate them to the resulting `TransportEvent::ConnectResult` in order to provide it in the `Lib3hServerProtocol::Connected` response

AC:
- There is a test that ensures the request_id is being returned in a ConnectResult when connection is successful
- There is a test for when there is no connection that ensures the failure result is also returned in the error

Issue #168, opened 2019-07-11
Have RealEngine track requests to Core


ddd-mtl:
As lib3h, I can know when a client does not respond in a timely manner so that I can respond appropriately (e.g. notify the client).

When RealEngine sends requests to Core (e.g. `HandleStoreEntryAspect`, any `Handle*` enum variant), it should keep track of it and expect some response from Core, even in the form of a `SuccessResult` or `FailureResult`.
create a rust equivalent of                    https://github.com/holochain/n3h/blob/master/lib/n3h-common/track.js

Maybe: extend Lib3hProtocol to have some sort of RequestTimeout or FailureResult interaction with client.

Tasks:
- [ ] SPIKE: Decide how to respond appropriately to a client response timeout - and how to update the API
- [ ] Implement result of spike

AC:
- There are test that ensure lib3h is  able to respond appropriately when encountering a client  response timeout

PR #237, opened 2019-08-06
track request_ids

neonphog:
## PR summary

handle requset_ids both for lib3h making requests of core, and core making connect requests of lib3h.

RESOLVES https://github.com/holochain/lib3h/issues/168
RESOLVES https://github.com/holochain/lib3h/issues/173

## Review checklist 

### DIFF
--- a/crates/lib3h/src/engine/real_engine.rs
+++ b/crates/lib3h/src/engine/real_engine.rs
@@ -391,100 +391,118 @@ impl<T: Transport, D: Dht> RealEngine<T, D> {
             }
             // Our request for the publish_list has returned
             Lib3hClientProtocol::HandleGetAuthoringEntryListResult(msg) => {
-                if !self.request_track.has(&msg.request_id) {
-                    error!("untracked HandleGetAuthoringEntryListResult");
-                } else {
-                    match self.request_track.remove(&msg.request_id) {
-                        Some(data) => match data {
-                            RealEngineTrackerData::GetAuthoringEntryList => (),
-                            _ => error!("bad track type HandleGetAuthoringEntryListResult"),
-                        },
-                        None => error!("bad track type HandleGetAuthoringEntryListResult"),
-                    };
-                }
-                let maybe_space = self.get_space_or_fail(
-                    &msg.space_address,
-                    &msg.provider_agent_id,
-                    &msg.request_id,
-                    None,
-                );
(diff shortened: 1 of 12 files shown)


crates/lib3h/src/transport_wss/mod.rs
```diff
@@ -265,7 +267,17 @@ impl<T: Read + Write + std::fmt::Debug> Transport for TransportWss<T> {
     /// this should be called frequently on the event loop
     /// looks for incoming messages or processes ping/pong/close events etc
     fn process(&mut self) -> TransportResult<(DidWork, Vec<TransportEvent>)> {
-        let did_work = self.priv_process_stream_sockets()?;
+        let mut did_work = false;
+
+        while !self.inbox.is_empty() {
+            let cmd = self.inbox.pop_front().expect("queue not empty");
```
struktured: Is there a thread safe way to do this instead, like in a monad? I know we are single threaded but if we ever switch to multithreading this is the kinda bug which will bite us.

neonphog: good call :+1: gogo `while let`:

```rust
while let Some(ref cmd) = self.inbox.pop_front() {
    did_work = true;
    self.serve_TransportCommand(cmd)?;
}
```

crates/lib3h/src/engine/real_engine.rs
```diff
@@ -277,6 +286,18 @@ impl<T: Transport, D: Dht> RealEngine<T, D> {
             //   - From GetAuthoringList      : Convert to DhtCommand::BroadcastEntry
             //   - From DHT EntryDataRequested: Convert to DhtCommand::EntryDataResponse
             Lib3hClientProtocol::HandleFetchEntryResult(msg) => {
+                let mut is_data_for_author_list = false;
+                if self.request_track.has(&msg.request_id) {
+                    match self.request_track.remove(&msg.request_id) {
```
struktured: So the pattern is after we respond to a transaction we throw away the request id, right? And because we don't have persistence, these ids are just cached locally for now. Seems good enough for time being.

neonphog: :+1: hopefully we won't have so many outstanding requests to core that we need them to me anywhere but just in memory... but we can totally evaluate that if it becomes an issue : )

PR #251, opened 2019-08-15
Gossip entries occassionally

zippy:
## PR summary

Adds gossiping of entries as well as peer data into the Mirror DHT.
Implemented by adding a GossipEntriesToRequest to the DHT protocol, and adding a `force` boolean to the BroadcastEntry command which is used to force the pushing of an entry that's allready in the book-keeping list and thus would otherwise be ignored by the usual BoradcastEntry path.

Adding this change makes the `agent_id_gets_gossiped_on_startup` scenario in the `lib3h-client-protocal-integration-transport-memory` branch in core pass.

## Review checklist 


struktured:
While this PR does make the agent id appear, it doesn't seem to fix any other tests. So its either incorrect or incomplete and requires further anaylsis. I suggest we wait a bit until @neonphog finishes several related fixes on deck before we merge this but open to more aggresive approaches.

PR #256, opened 2019-08-20
add compare to crypto system buf

zippy:
## PR summary

adds compare function to crypto-system Buffer trait, and copies over the implementation for sodium secure buffer and does a bad implementation for u8 buffer.


crates/crypto_api/src/buffer/buffer_vec_u8.rs
```diff
@@ -19,4 +19,21 @@ impl Buffer for Vec<u8> {
...
+        let l = self.len();
+        if l != b.len() {
+            return 1;
+        }
+        let a = self.read_lock();
+        let b = b.read_lock();
+        let mut i = 0;
+        while i < l {
+            if a[i] != b[i] {
+                return 1;
```
neonphog: We actually need this to return -1 if a < b when interpreted as little endian per https://download.libsodium.org/doc/helpers#comparing-large-numbers.

We could potentially use https://docs.rs/libc/0.2.63/libc/fn.memcmp.html (though I can't remember, we might need to reverse the result)

crates/crypto_api/src/buffer/buffer_vec_u8.rs
```diff
@@ -19,4 +19,21 @@ impl Buffer for Vec<u8> {
     fn set_no_access(&self) {}
     fn set_readable(&self) {}
     fn set_writable(&self) {}
+
+    fn compare(&mut self, b: &mut Box<dyn Buffer>) -> i32 {
```
neonphog: what do you think about doing this as a provided implementation in the Trait, like https://github.com/holochain/lib3h/blob/develop/crates/crypto_api/src/buffer/mod.rs#L59-L62 ?

zippy: done

crates/crypto_api/src/buffer/buffer_vec_u8.rs
zippy: code now compares values, but probably doesn't do the right thing on different length buffers, but at lease I added a test.  @neonphog can you confirm what the appropriate expected behavior of compare should be?

zippy:
@neonphog added yet another test to make things clear and I'm quite mystified by the sodium compare algorithm  data seems to show that:

[45, 0] < [0, 1]  and
[45] > [0, 1]

that seems weird to me.

ddd-mtl:
found the issue. It compares right to left, starting from a's size.

crates/sodium/src/crypto_system/secure_buffer.rs
```diff
@@ -133,6 +133,12 @@ impl Buffer for SecureBuffer {
         }
         *self.p.borrow_mut() = ProtectState::ReadWrite;
     }
+
+    fn compare(&mut self, b: &mut Box<dyn Buffer>) -> i32 {
+        let mut a = self.write_lock();
+        let mut b = b.write_lock();
+        unsafe { rust_sodium_sys::sodium_compare(raw_ptr_char!(a), raw_ptr_char!(b), a.len()) }
+    }
```
neonphog: I think this should all be immutable, no?

```rust
    fn compare(&self, b: &Box<dyn Buffer>) -> i32 {
        let mut a = self.read_lock();
        let mut b = b.read_lock();
        unsafe { rust_sodium_sys::sodium_compare(raw_ptr_char_immut!(a), raw_ptr_char_immut!(b), a.len()) }
    }
```

crates/crypto_api/src/buffer/mod.rs
```diff
@@ -73,6 +73,30 @@ pub trait Buffer: Send + std::fmt::Debug + Deref<Target = [u8]> + DerefMut<Targe
...
         Ok(())
     }
+
+    /// compare this buffer to another buffer
+    /// Return :
+    /// | if a > b; return 1
+    /// | if a < b; return -1
+    /// | if a == b; return 0
+    #[allow(clippy::borrowed_box)]
+    fn compare(&mut self, b: &mut Box<dyn Buffer>) -> i32 {
```
neonphog: see other comment, this should be immutable
```suggestion
    fn compare(&self, b: &Box<dyn Buffer>) -> i32 {
```

PR #272, opened 2019-08-30
Ghost actor transport tests

zippy:
## PR summary

Adds a sample test of how the GhostActor Transport protocol should be used.  The test is intended to be complete around using all the different arms of the Transport Protocol enums.

## Review checklist 

### DIFF
--- a/crates/lib3h/tests/transport.rs
+++ b/crates/lib3h/tests/transport.rs
@@ -18,12 +18,13 @@ use url::Url;
 pub struct Mockernet {
     bindings: HashMap<Url, Tube>,
     connections: HashMap<Url, HashSet<Url>>,
+    errors: Vec<(Url, String)>,
 }
 
 pub enum MockernetEvent {
     Connection { from: Url },
     Message { from: Url, payload: Vec<u8> },
-    //    Error(String),
+    Error(String),
 }
 
 // The Mockernet is a Series-of-Tubes, which is the technical term for the
@@ -46,6 +47,7 @@ impl Mockernet {
         Mockernet {
             bindings: HashMap::new(),
             connections: HashMap::new(),
+            errors: Vec::new(),
         }
     }
 
@@ -56,6 +58,15 @@ impl Mockernet {
         } else {
             self.bindings.insert(url, Tube::new());
             true
+        }
+    }
+
+    /// remove a binding, this is should trigger an error event
+    pub fn unbind(&mut self, url: Url) {
+        if self.bindings.contains_key(&url) {
+            self.bindings.remove(&url);
+            self.errors
+                .push((url.clone(), format!("{} has become unbound", &url)));
         }
     }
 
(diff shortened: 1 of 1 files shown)


crates/lib3h/tests/transport.rs
```diff
@@ -0,0 +1,430 @@
...
+pub enum MockernetEvent {
+    Connection { from: Url },
+    Message { from: Url, payload: Vec<u8> },
+    Error(String),
+}
+
+// The Mockernet is a Series-of-Tubes, which is the technical term for the
+// sets of crossbeam channels in the bindings that mockernet shuttles
+// data between.
+// The Mockernet simulates a
```
neonphog: ahh! such a cliffhanger

crates/lib3h/tests/transport.rs
```diff
@@ -0,0 +1,430 @@
...
+pub struct Tube {
+    sender: crossbeam_channel::Sender<(Url, Vec<u8>)>,
+    #[allow(dead_code)]
+    receiver: crossbeam_channel::Receiver<(Url, Vec<u8>)>,
```
neonphog: I assume this isn't still dead code, can we remove the annotation?

crates/lib3h/tests/transport.rs
```diff
@@ -0,0 +1,430 @@
...
+            if errors.len() == 0 {
+                return Err(format!("{} not bound", address));
+            }
+            return Ok(events);
+        }
+        let (from, payload) = maybe_binding
+            .unwrap()
+            .receiver
+            .try_recv()
+            .map_err(|e| format!("{:?}", e))?;
```
neonphog: Is this correct? This may work fine for the deterministic process calls below, but if we integrated @struktured 's process framework, perhaps here we'd want to loop until we get an `Err(Empty)` response, and not report errors for `Err(Empty)`?

crates/lib3h/tests/transport.rs
```diff
@@ -0,0 +1,430 @@
...
+    fn take_parent_endpoint(&mut self) -> Option<TransportActorParentEndpoint> {
+        std::mem::replace(&mut self.endpoint_parent, None)
+    }
+
+    fn process_concrete(&mut self) -> GhostResult<WorkWasDone> {
+        detach_run!(&mut self.endpoint_self, |es| es.process(self.as_any()))?;
+        for msg in self.endpoint_self.as_mut().drain_messages() {
+            self.handle_msg_from_parent(msg)?;
+        }
+        self.handle_events_from_mockernet();
```
neonphog: not to be pedantic ; ) but this doesn't seem like boilerplate.

crates/lib3h/tests/transport.rs
```diff
@@ -0,0 +1,430 @@
...
+    }
+
+    fn handle_events_from_mockernet(&mut self) {
+        let mut mockernet = MOCKERNET.write().unwrap();
+        let our_url = self.bound_url.as_ref().unwrap();
+        if let Ok(events) = mockernet.process_for(our_url.clone()) {
+            for e in events {
+                match e {
+                    MockernetEvent::Message { from, payload } => {
+                        detach_run!(self.endpoint_self, |s| s.publish(
```
neonphog: `publish` shouldn't require `detach_run!`, since we're not passing in `self`

crates/lib3h/tests/transport.rs
```diff
@@ -0,0 +1,430 @@
...
+                            RequestToParent::ReceivedData {
+                                address: from,
+                                payload
+                            }
+                        ));
+                    }
+                    MockernetEvent::Connection { from } => {
+                        detach_run!(self.endpoint_self, |s| s
+                            .publish(RequestToParent::IncomingConnection { address: from }));
```
neonphog: don't need `detach_run!`

crates/lib3h/tests/transport.rs
```diff
@@ -0,0 +1,430 @@
...
+                    }
+                    MockernetEvent::Error(err) => {
+                        detach_run!(self.endpoint_self, |s| s.publish(
```
neonphog: don't need `detach_run!`

crates/lib3h/tests/transport.rs
zippy: good call fixed.

PR #273, opened 2019-09-02
Ghost actor typed callbacks

struktured:
## PR summary

This PR introduces an unconstrained type parameter `UserData` to `GhostCallback` which replaces the first parameter of the callback function from `&mut dyn Any` to `&mut UserData`. The `GhostTracker` and `GhostTrackerEntry` now maintain a map of callbacks for a specific `UserData` type.

Calls to `bookmark` and `request` now correspondingly accept `UserData` which must be the same type as the callbacks are assuming. Thus there is no need to convert the user data to the correct type using `downcast_mut`. 

Thus, in short,

```Rust
 Box::new(|u, ctx, data| { match a.downcast_mut::<UserData>() ... })
```
becomes

```Rust
 Box::new(|u:&mut UserData, ctx, data| { // no need for downcast! })
```
 
and the `GhostParentContextEndpoint` now requires a `UserData` type parameter (and thus so does `as_context_endpoint` in `GhostEndpoint`.

## Review checklist 

### DIFF
--- a/crates/ghost_actor/src/ghost_actor.rs
+++ b/crates/ghost_actor/src/ghost_actor.rs
@@ -35,12 +35,12 @@ pub struct GhostParentWrapper<
 }
 
 impl<
-        Context,
+        Context: 'static,
         RequestToParent,
         RequestToParentResponse,
         RequestToChild,
-        RequestToChildResponse,
-        Error,
+        RequestToChildResponse: 'static,
+        Error: 'static,
     >
     GhostParentWrapper<
         Context,
@@ -77,12 +77,12 @@ impl<
     }
 
     /// see GhostContextEndpoint::request
-    pub fn request(
+    pub fn request<A: Any>(
         &mut self,
         timeout: std::time::Duration,
         context: Context,
         payload: RequestToChild,
-        cb: GhostCallback<Context, RequestToChildResponse, Error>,
+        cb: GhostCallback<Context, RequestToChildResponse, Error, A>,
     ) {
         self.endpoint.request(timeout, context, payload, cb)
     }
--- a/crates/ghost_actor/src/ghost_actor.rs
+++ b/crates/ghost_actor/src/ghost_actor.rs
@@ -423,7 +423,7 @@ mod tests {
         // now lets try posting a request with a callback which just saves the response
         // value to the parent's statee
         let cb: GhostCallback<FakeParent, TestContext, TestMsgInResponse, TestError> =
-            Box::new(|dyn_parent, _context, callback_data| {
+            Box::new(|parent, _context, callback_data| {
                 if let GhostCallbackData::Response(Ok(TestMsgInResponse(payload))) = callback_data {
                     dyn_parent.state = payload;
                 }
--- a/crates/ghost_actor/src/ghost_actor.rs
+++ b/crates/ghost_actor/src/ghost_actor.rs
@@ -82,7 +82,7 @@ impl<
         timeout: std::time::Duration,
         context: Context,
         payload: RequestToChild,
-        cb: GhostCallback<Context, RequestToChildResponse, Error, A>,
+        cb: GhostCallback<A, Context, RequestToChildResponse, Error>,
     ) {
         self.endpoint.request(timeout, context, payload, cb)
     }
--- a/crates/ghost_actor/src/ghost_actor.rs
+++ b/crates/ghost_actor/src/ghost_actor.rs
@@ -425,7 +425,7 @@ mod tests {
         let cb: GhostCallback<FakeParent, TestContext, TestMsgInResponse, TestError> =
             Box::new(|parent, _context, callback_data| {
                 if let GhostCallbackData::Response(Ok(TestMsgInResponse(payload))) = callback_data {
-                    dyn_parent.state = payload;
+                    parent.state = payload;
                 }
                 Ok(())
             });
--- a/crates/ghost_actor/src/ghost_actor.rs
+++ b/crates/ghost_actor/src/ghost_actor.rs
@@ -8,6 +8,7 @@ use std::any::Any;
 /// You only have to call process() on this one struct, and it provides
 /// all the request / drain_messages etc functions from GhostEndpoint.
 pub struct GhostParentWrapper<
+    UserData,
     Context,
     RequestToParent,
     RequestToParentResponse,
@@ -24,6 +25,7 @@ pub struct GhostParentWrapper<
 > {
     actor: Actor,
     endpoint: GhostContextEndpoint<
+        UserData,
         Context,
         RequestToChild,
         RequestToChildResponse,
@@ -34,6 +36,7 @@ pub struct GhostParentWrapper<
 }
 
 impl<
+        UserData: 'static,
         Context: 'static,
         RequestToParent,
         RequestToParentResponse,
@@ -49,6 +52,7 @@ impl<
         >,
     >
     GhostParentWrapper<
+        UserData,
         Context,
         RequestToParent,
         RequestToParentResponse,
@@ -73,12 +77,12 @@ impl<
     }
 
     /// see GhostContextEndpoint::request
-    pub fn request<A: Any>(
+    pub fn request(
         &mut self,
         timeout: std::time::Duration,
         context: Context,
         payload: RequestToChild,
-        cb: GhostCallback<A, Context, RequestToChildResponse, Error>,
+        cb: GhostCallback<UserData, Context, RequestToChildResponse, Error>,
     ) {
         self.endpoint.request(timeout, context, payload, cb)
     }
@@ -91,14 +95,15 @@ impl<
     }
 
     /// see GhostContextEndpoint::process and GhostActor::process
-    pub fn process(&mut self, actor: &mut dyn Any) -> GhostResult<()> {
+    pub fn process(&mut self, user_data: &mut UserData) -> GhostResult<()> {
         self.actor.process()?;
-        self.endpoint.process(actor)?;
+        self.endpoint.process(user_data)?;
         Ok(())
     }
 }
 
 impl<
+        UserData,
         Context,
         RequestToParent,
         RequestToParentResponse,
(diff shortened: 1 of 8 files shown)


crates/ghost_actor/src/ghost_actor.rs
```diff
@@ -422,13 +422,10 @@ mod tests {
 
         // now lets try posting a request with a callback which just saves the response
         // value to the parent's statee
-        let cb: GhostCallback<TestContext, TestMsgInResponse, TestError> =
+        let cb: GhostCallback<FakeParent, TestContext, TestMsgInResponse, TestError> =
             Box::new(|dyn_parent, _context, callback_data| {
```
zippy:
```suggestion
            Box::new(|parent, _context, callback_data| {
```

crates/ghost_actor/src/ghost_actor.rs
```diff
@@ -422,13 +422,10 @@ mod tests {
...
-                let mutable_parent = dyn_parent
-                    .downcast_mut::<FakeParent>()
-                    .expect("should be a FakeParent");
                 if let GhostCallbackData::Response(Ok(TestMsgInResponse(payload))) = callback_data {
-                    mutable_parent.state = payload;
+                    dyn_parent.state = payload;
```
zippy:
```suggestion
                    parent.state = payload;
```

crates/ghost_actor/src/ghost_channel.rs
```diff
@@ -414,12 +420,9 @@ mod tests {
         // this genrates a callback for requests that simply puts the callbackdata  into
         // the FakeActor's state String, thus for testing we can just look in the actors's
         // state to see if the callback was run.
-        fn cb_factory() -> GhostCallback<TestContext, TestMsgOutResponse, TestError> {
+        fn cb_factory() -> GhostCallback<FakeActor, TestContext, TestMsgOutResponse, TestError> {
             Box::new(|dyn_me, _context, callback_data| {
```
zippy:
```suggestion
            Box::new(|me, _context, callback_data| {
```

crates/ghost_actor/src/ghost_channel.rs
```diff
@@ -414,12 +420,9 @@ mod tests {
...
-                let mutable_me = dyn_me
-                    .downcast_mut::<FakeActor>()
-                    .expect("should be a our fake actor which is just a String");
-                mutable_me.0 = format!("{:?}", callback_data);
+                dyn_me.0 = format!("{:?}", callback_data);
```
zippy:
```suggestion
                me.0 = format!("{:?}", callback_data);
```

crates/ghost_actor/src/ghost_tracker.rs
```diff
@@ -140,17 +147,13 @@ mod tests {
         let mut actor = TestTrackingActor::new("test_request_id_prefix");
         let context = TestContext("some_context_data".into());
 
-        let cb: GhostCallback<TestContext, TestCallbackData, TestError> =
+        let cb: GhostCallback<TestTrackingActor, TestContext, TestCallbackData, TestError> =
             Box::new(|dyn_me, context, callback_data| {
```
zippy:
```suggestion
            Box::new(|me, context, callback_data| {
```

crates/ghost_actor/src/ghost_tracker.rs
```diff
@@ -140,17 +147,13 @@ mod tests {
...
-                let mutable_me = dyn_me
-                    .downcast_mut::<TestTrackingActor>()
-                    .expect("should be a TestTrakingActor");
-
                 // and we'll check that we got our context back too because we
                 // might have used it to determine what to do here.
                 assert_eq!(context.0, "some_context_data");
                 if let GhostCallbackData::Response(Ok(TestCallbackData(payload))) = callback_data {
-                    mutable_me.state = payload;
+                    dyn_me.state = payload;
```
zippy:
```suggestion
                    me.state = payload;
```

crates/ghost_actor/src/ghost_tracker.rs
```diff
@@ -198,16 +201,12 @@ mod tests {
     fn test_ghost_tracker_should_timeout() {
         let mut actor = TestTrackingActor::new("test_request_id_prefix");
         let context = TestContext("foo".into());
-        let cb: GhostCallback<TestContext, TestCallbackData, TestError> =
+        let cb: GhostCallback<TestTrackingActor, TestContext, TestCallbackData, TestError> =
             Box::new(|dyn_me, _context, callback_data| {
```
zippy:
```suggestion
            Box::new(|me, _context, callback_data| {
```

crates/ghost_actor/src/ghost_tracker.rs
```diff
@@ -198,16 +201,12 @@ mod tests {
...
-                let mutable_me = dyn_me
-                    .downcast_mut::<TestTrackingActor>()
-                    .expect("should be a TestTrakingActor");
-
                 // when the timeout happens the callback should get
                 // the timeout enum in the callback_data
                 match callback_data {
-                    GhostCallbackData::Timeout => mutable_me.state = "timed_out".into(),
+                    GhostCallbackData::Timeout => dyn_me.state = "timed_out".into(),
```
zippy:
```suggestion
                    GhostCallbackData::Timeout => me.state = "timed_out".into(),
```

crates/ghost_actor/src/ghost_tracker.rs
```diff
@@ -70,13 +70,20 @@ impl<Context, CbData, E> GhostTracker<Context, CbData, E> {
...
-        cb: GhostCallback<Context, CbData, E>,
+        cb: GhostCallback<A, Context, CbData, E>,
     ) -> RequestId {
         let request_id = RequestId::with_prefix(&self.request_id_prefix);
+
+        let cb: GhostCallback<dyn Any, _, _, _> = Box::new(move |a, ctx, data| {
+            let a = a
+                .downcast_mut::<A>()
+                .expect("downcast Any to specific actor A");
+            (cb)(a, ctx, data)
```
neonphog: :+1:

crates/ghost_actor/src/ghost_actor.rs
```diff
@@ -226,12 +226,12 @@ pub struct GhostParentWrapperDyn<
 }
 
 impl<
-        Context,
+        Context: 'static,
```
neonphog: The consequence of this is that we wouldn't be able to store any non-static references in context enums... which may be ok since they have access to the "user data" for any reference type stuff?

crates/ghost_actor/src/ghost_actor.rs
```diff
@@ -226,12 +226,12 @@ pub struct GhostParentWrapperDyn<
...
         RequestToParent,
         RequestToParentResponse,
         RequestToChild,
-        RequestToChildResponse,
-        Error,
+        RequestToChildResponse: 'static,
+        Error: 'static,
```
neonphog: These two should be fine... we should probably add `'static' to all the Request* types, because we don't want references in those or the error type.

crates/ghost_actor/src/ghost_actor.rs
```diff
@@ -73,12 +73,12 @@ impl<
     }
 
     /// see GhostContextEndpoint::request
-    pub fn request(
+    pub fn request<A: Any>(
```
neonphog: Do we want to change this to something like `UserData: Any`?

crates/ghost_actor/src/ghost_actor.rs
struktured: @neonphog I'm ok with this limitation but would be interested in eliminating the static constraint if possible in the future.

crates/ghost_actor/src/ghost_actor.rs
struktured: @neonphog why don't we want references in request types? Because we need to serialize them,  or share among threads?

crates/ghost_actor/src/ghost_actor.rs
struktured: Done in https://github.com/holochain/lib3h/pull/273/commits/ee7046d0a9c9a447067b4a24c4b14691089de164

struktured:
> Woot, this'll be much more ergonomic / typesafe. Esp if we can extend it to the `process()` function

Wish granted: https://github.com/holochain/lib3h/pull/273/commits/ee7046d0a9c9a447067b4a24c4b14691089de164

crates/ghost_actor/src/ghost_tracker.rs
```diff
@@ -17,25 +17,27 @@ pub enum GhostCallbackData<CbData, E> {
 /// if you want to mutate the state of a struct instance, pass it in
 /// with the `handle` or `process` call.
 /// (see detach crate for help with self refs)
-pub type GhostCallback<Context, CbData, E> =
-    Box<dyn Fn(&mut dyn Any, Context, GhostCallbackData<CbData, E>) -> GhostResult<()> + 'static>;
+pub type GhostCallback<UserData, Context, CbData, E> =
+    Box<dyn Fn(&mut UserData, Context, GhostCallbackData<CbData, E>) -> GhostResult<()> + 'static>;
```
neonphog: dude! what the heck? It looks like you just switched the type and it magically started working, lol. I wonder what I was doing wrong when I tried this : )

crates/ghost_actor/src/ghost_actor.rs
neonphog: Exactly, If we do end up wanting to upgrade this to a "managed" actor framework where we don't have to manually call process on each child, we'd need these messages to be passable without references.

crates/lib3h/src/transport/protocol.rs
```diff
@@ -68,14 +68,18 @@ pub type DynTransportActor = Box<
         TransportError,
     >,
 >;
+
+//type UserData = TransportEncoding;
```
zippy: I don't think this comment should be here...

crates/ghost_actor/src/ghost_tracker.rs
struktured: @neonphog I think the static constraints might have been important.

crates/lib3h/src/transport/protocol.rs
struktured: Removed.

PR #275, opened 2019-09-03
SimChat ghost actor part 1

willemolding:
## PR summary

- Renames half-busy-chat (HBC) to simchat
- Updates the skeleton simchat to speak ghost protocol to the engine
- Can connect to a network passing a peer URI (engine currently times out in this request)
- Adds ability to join a space (once engine supports)
- Adds ability to leave a space (once engine supports)
- Adds ability to direct message an agent (assuming you can know their agent_id somehow and the engine supports)

## Review checklist 


crates/tools/sim_chat/src/main.rs
```diff
@@ -0,0 +1,164 @@
...
+        .expect("failed to set linefeed prompt");
+
+    let rl_t = rl.clone();
+    let mut cli = lib3h_sim_chat::SimChat::new(
+        engine_builder,
+        Box::new(move |event| {
+            match event {
+                ChatEvent::JoinSuccess { channel_id, .. } => {
+                    rl_t.set_prompt(&format!("#{}> ", channel_id).to_string())
+                        .expect("failed to set linefeed prompt");
```
neonphog: Should we have a channel HashSet? It might be cool to let them join any number of channels, and if they join one they've previously already join, it just promotes that one to the prompt (i.e. the one outgoing messages will be sent on), and they will see incoming messages on all channels.

alternately, when they do join and are already part of a channel we could automatically send a LeaveSpace

crates/tools/sim_chat/src/main.rs
```diff
@@ -0,0 +1,164 @@
...
+                }
+                ChatEvent::PartSuccess => {
+                    rl_t.set_prompt("SimChat> ")
```
neonphog: see previous comment. If we want to support multiple channel membership, we might want to check the HashSet and promote another channel to the prompt.

crates/tools/sim_chat/src/lib.rs
```diff
@@ -0,0 +1,513 @@
...
+                while thread_continue_inner.load(Ordering::Relaxed) {
+                    // call process to make stuff happen
+                    parent_endpoint.process(&mut ()).unwrap();
+                    engine.process().unwrap();
+
+                    // gather all the ChatEvents
+                    // Receive directly from the crossbeam channel
+                    // and convert relevent N3H protocol messages to chat events
+                    let direct_chat_events = out_recv.try_iter();
+                    let n3h_chat_events = parent_endpoint.drain_messages().into_iter().filter_map(
```
neonphog: may not want to use `n3h`, lest we confuse people : )

PR #276, opened 2019-09-03
Ghost Actor Default Timeouts

neonphog:
## PR summary

This creates a default timeout that can be used for most normal use-cases.

Adds builder pattern for GhostTracker, and ContextEndpoints + adds default timeout to both which defaults to 2 seconds.

Separates:

- `bookmark_timeout` from `bookmark` (which uses the default timeout)
- `request_timeout` from `request` (which uses the default timeout)

## Review checklist 

### DIFF
--- a/crates/ghost_actor/src/ghost_actor.rs
+++ b/crates/ghost_actor/src/ghost_actor.rs
@@ -1,6 +1,4 @@
-use crate::{
-    GhostCallback, GhostContextEndpoint, GhostEndpoint, GhostMessage, GhostResult, WorkWasDone,
-};
+use crate::prelude::*;
 
 /// helper struct that merges (on the parent side) the actual child
 /// GhostActor instance, with the child's ghost channel endpoint.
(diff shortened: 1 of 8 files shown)


crates/ghost_actor/src/ghost_actor.rs
```diff
@@ -270,12 +273,21 @@ impl<
...
-        timeout: std::time::Duration,
         context: Context,
         payload: RequestToChild,
         cb: GhostCallback<Context, RequestToChildResponse, Error>,
     ) {
-        self.endpoint.request(timeout, context, payload, cb)
+        self.endpoint.request(context, payload, cb)
+    }
+
+    pub fn request_timeout(
```
struktured: Just an idea, not blocker- we could future proof this by making an optional argument container, much like holochain does for various functions. For instance here's the simple `get_entry` invoking the more advanced version:

```Rust
pub fn get_entry(address: &Address) -> ZomeApiResult<Option<Entry>> {
    let entry_result = get_entry_result(address, GetEntryOptions::default())?;
...
}
```

crates/ghost_actor/src/ghost_tracker.rs
```diff
@@ -28,22 +28,50 @@ struct GhostTrackerEntry<Context, CbData, E> {
...
 
-impl<Context, CbData, E> GhostTracker<Context, CbData, E> {
-    /// create a new tracker instance (with request_id prefix)
-    pub fn new(request_id_prefix: &str) -> Self {
+impl Default for GhostTrackerBuilder {
+    fn default() -> Self {
         Self {
-            request_id_prefix: request_id_prefix.to_string(),
+            request_id_prefix: "".to_string(),
+            default_timeout: std::time::Duration::from_millis(2000),
```
struktured: maybe make a constant for this timeout value, just to make it slightly easier to dig up when debugging timeout conditions. Although perhaps the `Default` trait encapsulates it enough already.

crates/ghost_actor/src/ghost_actor.rs
neonphog: :+1: I like that. Updated.

crates/ghost_actor/src/ghost_tracker.rs
neonphog: :+1:

PR #277, opened 2019-09-04
Ghost actor dht

ddd-mtl:
## PR summary

Implemented GhostActor for MirrorDht. Created a DhtActor and protocol.
rrdht disabled for now.
Removed Dht trait. Renamed file to dht_config.
Temporary outbox in RealEngine until we have engine & gateway actor.
Some other shortcuts have been taken in engine and gateway by taking advantage that its a fullsync dht for now.
P2pGateway now has helper functions for synchronous calls to its child dht actor.
Moved GatewayWrapper to its own file.

Had to change `connection_id_list()` to use `&mut self` because internally we are making a request which mutates.

Test behavior change:
When Lib3hClient sends a HoldEntry, we do not expect a HandleFetchEntry response since the data is already there.


crates/lib3h/src/dht/dht_protocol.rs
```diff
@@ -2,29 +2,109 @@ use crate::dht::PeerAddress;
...
 
-#[derive(Debug, PartialEq, Clone)]
-pub enum DhtEvent {
+#[derive(Debug)]
+pub enum DhtRequestToChildResponse {
+    RequestPeer(Option<PeerData>),
+    RequestPeerList(Vec<PeerData>),
+    RequestThisPeer(PeerData),
+    RequestEntryAddressList(Vec<Address>),
+    RequestAspectsOf(Option<Vec<Address>>),
```
struktured: What is difference between `None` and `Some(vec![])` here?

crates/lib3h/src/dht/dht_protocol.rs
```diff
@@ -2,29 +2,109 @@ use crate::dht::PeerAddress;
...
     BroadcastEntry(EntryData),
-    /// Owner notifies us that is is not holding an entry anymore.
+    /// Parent notifies us that is is not holding an entry anymore.
     DropEntryAddress(Address),
-    /// Owner's response to ProvideEntry request
-    EntryDataResponse(FetchDhtEntryResponseData),
+
+    /// Requests
+    /// Parent wants PeerData for a specific Peer
+    RequestPeer(String),
```
struktured: Can do we better than a `String` type here?

crates/lib3h/src/dht/mirror_dht.rs
```diff
@@ -301,18 +282,46 @@ impl MirrorDht {
...
+    fn take_parent_endpoint(&mut self) -> Option<DhtEndpoint> {
+        std::mem::replace(&mut self.endpoint_parent, None)
+    }
+
+    fn process_concrete(&mut self) -> GhostResult<WorkWasDone> {
+        detach_run!(&mut self.endpoint_self, |es| es.process(&mut ()))?;
+        for request in self.endpoint_self.as_mut().drain_messages() {
+            //debug!("@MirrorDht@ serving request: {:?}", request);
+            self.handle_request_from_parent(request)
+                .expect("no ghost errors");
```
struktured: I wonder if @neonphog has an example of handling this with a result combinator. I think using the [collect()](https://doc.rust-lang.org/rust-by-example/error/iter_result.html#fail-the-entire-operation-with-collect) trick would be appropriate here.

crates/lib3h/src/dht/mirror_dht.rs
```diff
@@ -301,18 +282,46 @@ impl MirrorDht {
...
-    #[allow(non_snake_case)]
-    fn serve_DhtCommand(&mut self, cmd: &DhtCommand) -> Lib3hResult<Vec<DhtEvent>> {
-        debug!("@MirrorDht@ serving cmd: {:?}", cmd);
-        // Note: use same order as the enum
-        match cmd {
+impl MirrorDht {
+    #[allow(irrefutable_let_patterns)]
+    fn handle_request_from_parent(&mut self, mut request: DhtToChildMessage) -> Lib3hResult<()> {
+        debug!("@MirrorDht@ serving request: {:?}", request);
+        match request.take_message().expect("exists") {
```
struktured: I think this api is in flux, but if it compiles now, then it's good enough for me.

crates/lib3h/src/dht/mirror_dht.rs
```diff
@@ -326,44 +335,43 @@ impl MirrorDht {
...
+                            None => {
+                                self.endpoint_self
+                                    .publish(DhtRequestToParent::HoldPeerRequested(
+                                        gossiped_peer.clone(),
+                                    ));
+                            }
+                            Some(known_peer) => {
+                                // Update Peer timestamp
+                                if gossiped_peer.timestamp > known_peer.timestamp {
+                                    let _ = self.add_peer(&gossiped_peer);
```
struktured: It's not obvious why the return value to `add_peer` is ignored here and in a few other places. maybe just give the unused variable it a meaningful name or document it.

crates/lib3h/src/dht/mirror_dht.rs
```diff
@@ -383,67 +391,112 @@ impl MirrorDht {
...
+                            _ => panic!("bad context"),
+                        };
+                        if let DhtRequestToParentResponse::RequestEntry(entry_response) = response {
+                            println!("4. In DhtRequestToChild::RequestEntry Responding...");
+                            let payload =
+                                Ok(DhtRequestToChildResponse::RequestEntry(entry_response));
+                            request.respond(payload);
+                        } else {
+                            panic!("bad response to RequestEntry: {:?}", response);
+                        }
```
struktured: Lots of `panic!`s in this code- is this the type of error handling we want here?

crates/lib3h/src/dht/dht_protocol.rs
ddd-mtl: `None` means there is no entry at all.
`Some(vec![]` means there is an entry with no aspects.

crates/lib3h/src/dht/dht_protocol.rs
ddd-mtl: Not really since its the typed used in the DHT.

crates/lib3h/src/dht/mirror_dht.rs
ddd-mtl: The return value tells us if the added content is new or just an update.
Here we already know its known content so we don't care about this return value.

crates/lib3h/src/dht/mirror_dht.rs
ddd-mtl: Probably not, but I'm guessing we are going to macroize this in some cleaner way anyway so taking the short route here.

PR #282, opened 2019-09-04
RFC: Tuple for drain_messages

neonphog:
## PR summary

I *just* updated the ghost_actor crate for now... If we're liking how this is working, I can update the rest of our monorepo.

## Review checklist 

### DIFF
--- a/crates/ghost_actor/src/ghost_actor.rs
+++ b/crates/ghost_actor/src/ghost_actor.rs
@@ -136,7 +136,7 @@ impl<
     /// see GhostContextEndpoint::drain_messages
     fn drain_messages(
         &mut self,
-    ) -> Vec<GhostMessage<RequestToParent, RequestToChild, RequestToParentResponse, Error>> {
+    ) -> Vec<(RequestToParent, GhostResponder<RequestToParentResponse, Error>)> {
         self.endpoint.drain_messages()
     }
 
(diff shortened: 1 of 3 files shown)


crates/ghost_actor/src/ghost_channel.rs
```diff
@@ -452,11 +456,26 @@ impl<
                         request_id,
                         payload,
                     } => {
+                        let responder: GhostResponder<RequestToSelfResponse, Error> = match request_id {
```
zippy: So it looks like this makes the `respond()` method on GhostMessage obsolete? Maybe we should move this code into the GhostMessage and convert `respond()` into `make_responder()`?

crates/ghost_actor/src/ghost_channel.rs
```diff
@@ -452,11 +456,26 @@ impl<
...
+                            Some(request_id) => {
+                                let sender = self.sender.clone();
+                                Box::new(move |payload| {
+                                    sender.send(GhostEndpointMessage::Response {
+                                        request_id,
+                                        payload,
+                                    }).expect("should send");
+                                })
+                            }
+                            None => Box::new(|_|{}),
```
zippy: This kind of worries me.  At the very least it should log an unexpected code path, maybe panic?

crates/ghost_actor/src/ghost_actor.rs
```diff
@@ -436,14 +465,13 @@ mod tests {
 
             // In this test actor we simply take all the messages we get and
             // add them to our internal state.
-            for mut msg in self.endpoint_as_child.as_mut().drain_messages() {
-                let payload = match msg.take_message().expect("exists") {
+            for (payload, responder) in self.endpoint_as_child.as_mut().drain_messages() {
```
zippy: I think in practice we should probably still call what's drained a `message`, not a `payload`.

crates/ghost_actor/src/ghost_actor.rs
```diff
@@ -436,14 +465,13 @@ mod tests {
...
+                let payload = match payload {
                     TestMsgIn(payload) => payload,
                 };
                 self.internal_state.push(payload.clone());
-                if msg.is_request() {
-                    msg.respond(Ok(TestMsgInResponse(format!("we got: {}", payload))))
-                };
+                responder(
```
zippy: I don't like how the possibility of messages being events not requests becomes obscured here...

ddd-mtl:
I don't like this as it adds yet another concept and don't see the value gain.
Maybe rename `take_message()` to `take_payload()` because I find `msg.take_message()` a bit confusing.

PR #283, opened 2019-09-05
seeded processing in assert_processed!

thedavidmeister:
## PR summary

this hooks into the existing processing macros to make ABABABABABABA look more like ABBABABAAABABABBBAA (pseudo random)

thanks to @ddd-mtl for firstly making tests that _have_ a `.process()` call at all and @struktured for wrapping it up in macros that we can abstract it all away with to make this PR super easy :)

the idea is to get deterministic behaviour that more closely mimics non-deterministic behaviour

the assumption here is that A and B both happen 50% of the time, we could get increasingly sophisticated over time with the unit generator e.g. "B processes twice as fast as A" or "throw a normal distribution across these 50 engines"

the reason i am using a specific algo rather than just whatever rust provides is to make sure that we're working with something that we can audit/port/reproduce/etc.

it's seeded and prints out the seed used alongside the existing info that was there

basically, if something breaks due to the processing order changing then paste the seed straight into the macro to reproduce

in the future we could do something more sophisticated like passing in seeds as CLI flags or environment config or something... but for now just do it manually

the specific algo chosen is xoroshiro128+ which is fast and passes almost all statistical quality checks with a workaround for the known weakness (use the high bits for boolean as i'm doing here) and the period is easily large enough for our needs, and it's implemented across a variety of languages which means the sequences are portable if we ever find ourselves needing to reproduce seeds somewhere else

it also allows us to "jump ahead" if we ever need to, effectively spawning many PRNGs from a single seed (it's the equivalent to calling it 2^64 times in a single hop) and we can do _that_ 2^64 times before looping back to the start for any given seed

i exposed it as a unit prng (between 0 and 1) and boolean with the expectation that in the future we may be switching between more than 2 options, so we'd want a float then to distribute across N engines

i seeded it all with the default rust behaviour for getting a random u64, which IIRC just dials out to the OS-specific source of randomness - if we think this isn't good enough for some reason there is the option to pass a single u64 to the "split mix" algorithm that is designed to be a useful way to generate a seed for xoroshiro (by being a prng algorithm that is sufficiently not correlated to the "main" xoroshiro algo), tradeoff: we end up with u64 different seeds instead of u128 different seeds

the rust crate is maintained by an old colleage of mine, i'm sure we could hit him up if we needed anything, and i've done an implementation of it too in clojure (JVM) and clojurescript (JS), so i think we have a lot of room to move if we need something to work a certain way (i'm not aware of anything atm, just FYI)

also FYI the name xoroshiro128+ is basically what it does under the hood... it does XOr > ROtate > SHIft > ROtate and is addition of 128 bits of internal state

also FYI this family of prngs is used by browsers e.g. chrome and firefox under the hood, but it's an older version of the algo from the same researchers last time i checked

image

image

## Review checklist 


thedavidmeister:
interestingly this build https://ci.appveyor.com/project/thedavidmeister/lib3h/builds/27221406 failed!!

i can't reproduce it though because i was reseeding on every macro call, so i've moved to a single lazy static which allows for a single seed, which allows for reproducing errors, which is the whole point :)

thedavidmeister:
:thinking: 

https://circleci.com/gh/holochain/lib3h/1415?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link fails

https://circleci.com/gh/holochain/lib3h/1417?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link passes

same seed??

PR #305, opened 2019-09-06
Ghost actor bootstrap connect

struktured:
## PR summary

This PR manually merges the `bootstrap-connect` branch which incorporates `Opaque`, bootstrap connecting logic, some ~~memory server leak fixes~~, and other small improvements.

## Review checklist 


crates/lib3h_protocol/src/data_types.rs
```diff
@@ -6,6 +6,61 @@ use url::Url;
 /// (entry_address, aspect_address)
 pub type AspectKey = (Address, Address);
 
+/// Represents an opaque vector of bytes. Lib3h will
+/// store or transfer this data but will never inspect
+/// or interpret its contents
+#[derive(Clone, Eq, PartialEq, Deserialize, Serialize)]
+pub struct Opaque(#[serde(with = "base64")] Vec<u8>);
```
neonphog: should we shrinkwraprs, since we're pulling that in as a dependency for ghost_actor in any case? - https://github.com/holochain/lib3h/blob/ghost-actor/crates/ghost_actor/src/lib.rs#L25-L26

crates/lib3h/src/transport/memory_mock/ghost_transport_memory.rs
```diff
@@ -126,18 +139,24 @@ impl
...
+                                                            msg.respond(Err(TransportError::new(format!(
+                                                                "No Memory server at this address: {}",
+                                                                my_addr
+                                                            ))))?;
+                                                            continue;
+                                                        }
+
+                                                        let server_ref = maybe_server.unwrap();
+                                                        let mut server = server_ref.get();
+                            */
```
neonphog: should we delete this block?

crates/lib3h_protocol/src/data_types.rs
struktured: That's a good idea. Will look into it for this PR or subsequent one depending on timing.

;
```diff
@@ -0,0 +1,452 @@
+#[macro_use]
```
neonphog: You have a `;` file here

PR #310, opened 2019-09-07
[WIP] GhostActor tracing

maackle:
## PR summary

Demonstrates how to integrate tracing in lib3h. GhostActor users can pass in a rustracing Span, 

### TODO

- [ ] Maybe add a span parameter to `GhostCallback` so that the span can be continued, or followed from, the span that led up to the request being handled
    - [ ] examine whether it's really necessary to pass Context into the callback function, since it seems to be available at the moment the callback closure is created, so it could just be moved into the closure at that point
- [ ] See how GhostActor is actually being used
- [ ] Come up with an initial set of trace points, logging events with sufficient detail, within and without GhostActor
- [ ] Verify that Spans aren't getting ended when they are moved between scopes
- [ ] Set up a Jaeger Reporter

## Review checklist 


maackle:
This was part of my initial explorations, and things have changed since then. And I also forgot to make this a "Draft". Starting over with a fresh PR.

PR #304, opened 2019-09-05
GhostEngine

zippy:
## PR summary

This PR gets part way to integrating the ghost actor gateway and multiplexer into a ghost engine. Things still to do to complete the integration:
- handle the three sub-process calls network/space/ugly in  `GhostEngine` `process_concrete()`
- fix `handle_HandleGetGossipingEntryListResult`
- fix `handle_HandleGetAuthoringEntryListResult`
- fix `shutdown`
- make the tests actually observe that the correct messages are posted to the right places
- fix non-determinism in ghost_memory_transport test.

## Review checklist 
- [x] The story has unit or integration tests
- [x] No new bugs, and any tech-debt is identified and justified
- [x] There is enough API documentation (how to use)
- [x] There is enough code documentation (how the code works)

PR #311, opened 2019-09-09
Ghost context refactor

maackle:
## PR summary

- Removes `context` param from `GhostCallback` fn, moving data into closures directly as necessarily
- Makes `GhostCallback` FnOnce instead of Fn
- Adds type constraint to all `Context` generics, requiring that it implement `GhostContext`, which is a trait that exposes a tracing Span (currently stubbed as a Fn<() -> ()>)
- impl GhostContext where necessary (with stubs)
- Add `GhostMessageData`, which stores everything in the `GhostMessage` except for the Sender
    - This is because `Lib3hError` is not `Clone` because of `io::Error`. 
    - Used to get the `message` into closures with minimal hassle. Could probably do without this struct with a more surgical refactor.

## Review checklist 


crates/lib3h/src/engine/ghost_engine_wrapper.rs
```diff
@@ -68,32 +68,32 @@ where
...
 
-    fn make_callback() -> GhostCallback<
-        LegacyLib3h<Engine, EngineError>,
-        ClientRequestContext,
-        ClientToLib3hResponse,
-        EngineError,
-    > {
+    fn make_callback(
+        request_id: String,
+    ) -> GhostCallback<LegacyLib3h<Engine, EngineError>, ClientToLib3hResponse, EngineError> {
```
maackle: Is it ok to pass request_id in as a param like this? This is the only data needed from the ClientRequestContext, so we need this here to move it into the callback closure now that context is removed

crates/tracing/src/lib.rs
```diff
@@ -0,0 +1,5 @@
+extern crate crossbeam_channel;
+extern crate rustracing;
+extern crate rustracing_jaeger;
+
+pub type Span = rustracing_jaeger::Span;
```
neonphog: Should we `shrinkwraprs` this so we can add attach custom helper functions / converters / etc?

crates/ghost_actor/Cargo.toml
```diff
@@ -14,6 +14,7 @@ repository = "https://github.com/holochain/lib3h"
 [dependencies]
 crossbeam-channel = "=0.3.8"
 detach = { version = "=0.0.10", path = "../detach" }
+lib3h_tracing = { version = "=0.0.1", path = "../tracing" }
```
neonphog: should this be lib3h tracing or should it be something more generic that can be used by core too?

maackle: @neonphog I figure this can move out of lib3h and become `holochain_tracing` when it's time. Or lib3h_tracing could even just be a wrapper around the more general one. Either way there's not much to it at this point, and probably never will be, so the move should be easy

neonphog: :+1:

crates/tracing/src/lib.rs
maackle: Cool I didn't know about shrinkwrap. Probably! For now it's just a stub that's not used for real.

crates/lib3h/src/engine/ghost_engine_wrapper.rs
neonphog: I'm ok with this :+1:

crates/tracing/src/lib.rs
neonphog: :+1:

maackle:
@neonphog 

> You may indeed need access to a span from within the callback

I was thinking that the span would just get created and moved into the callback closure at the time it's created, as part of the request registration

> I think two things are getting conflated here

Yeah that may be the case. Maybe I should start implementing against this on a separate branch to see how it all pans out. It's hard to know the places we'll want access to the span up front. But I figure "everywhere" is close to accurate, so that's why I figured I'd just bolt it onto the context.

By the way, now that context isn't involved in functions, does that mean we can remove the context param from `request` in `GhostCanTrack` and everywhere else? Then the "non-tracing context" is really pretty arbitrary the way it's structured.

maackle:
@neonphog please take another look. I can also rename things however it makes sense (or maybe wait until we start using it to see what more natural name falls out).

maackle:
Also @zippy I'm working based on a branch that you're committing directly to so there are lots of merge conflicts. I'm OK to fix them at the end but please be aware of this context refactor.

maackle:
@neonphog yeah this is really just laying the groundwork for some real tracing stuff to come.

@struktured only a vague idea. It is probably overkill but it was easy to hijack the Context type param which is now being removed with this new TraceContext, so I figured why not. 1 + 1 - 1 = 1. If it gets onerous later then we can take away the type parameter.

maackle:
**NB:** I originally based this work on `ghost-engine`, #304. In order to not clobber the diff for that PR, I added a commit (`c04dff7`) which deletes all of Eric's work. We have to remember to revert that before merging 304.

PR #312, opened 2019-09-09
Transport Legacy to Ghost and back to Legacy

neonphog:
## PR summary

Gives me a bind point where I can inject the multiplexer. This will mostly all be deleted once we complete the integration, but there are some lifetime fixes in here that we will keep.

## Review checklist 


crates/lib3h/src/gateway/gateway_transport.rs
```diff
@@ -21,6 +21,9 @@ impl<'gateway> Transport for P2pGateway<'gateway> {
     // TODO #176 - Return a higher-level uri instead?
     fn connect(&mut self, uri: &Url) -> TransportResult<ConnectionId> {
         trace!("({}).connect() {}", self.identifier, uri);
+        if &self.identifier != "__network__" {
```
maackle: had to look up this magic string, could you import `NETWORK_GATEWAY_ID` instead?

neonphog: aw, crap... that's debugging that shouldn't be in there... we don't want to `std::process::exit(127)` ; )

neonphog: removed

neonphog:
closing for now, will wait and impl multiplexer directly between ghost gateway and ghost engine

PR #315, opened 2019-09-10
WIP - Ghost gateway actor 2nd attempt

ddd-mtl:
## PR summary

Removed Transport trait and ConnectionId


crates/lib3h/src/engine/mod.rs
```diff
@@ -90,27 +90,32 @@ pub struct TransportKeys {
 }
 
 /// Lib3h's 'real mode' as a NetworkEngine
-pub struct RealEngine<'engine> {
+pub struct RealEngine {
```
neonphog: I'm a little hesitant to remove explicit lifetimes, because they may need to come back in the future, and it's a lot of code shuffle to remove them / add them back. You can add a `phantom_lifetime: std::marker::PhantomData<&'engine i8>,` to the struct member list if the compiler is complaining about it not being used.

Although, in this case, real engine is moving over to ghost engine, so we can probably leave as-is.

crates/lib3h/src/engine/real_engine.rs
```diff
@@ -101,48 +99,88 @@ impl<'engine> RealEngine<'engine> {
...
+                };
+                if let transport::protocol::RequestToChildResponse::Bind(bind_data) = response {
+                    ud.binding = bind_data.bound_url;
+                } else {
+                    panic!("bad response to bind: {:?}", response);
+                }
+                Ok(())
+            }),
+        );
+        let _ = memory_transport.process();
```
neonphog: We should `?` this, or at the very least `unwrap()` rather than ignoring errors

crates/lib3h/src/engine/real_engine.rs
```diff
@@ -101,48 +99,88 @@ impl<'engine> RealEngine<'engine> {
...
+        let _ = memory_network_endpoint.process(&mut gateway_ud);
```
neonphog: ditto `?`

crates/lib3h/src/engine/real_engine.rs
```diff
@@ -101,48 +99,88 @@ impl<'engine> RealEngine<'engine> {
...
+            memory_transport
+                .take_parent_endpoint()
+                .expect("exists")
+                .as_context_endpoint_builder()
+                .request_id_prefix("tmem_to_child_")
+                .build::<GatewayUserData, Lib3hTrace>(),
+        );
+
+        // Bind & create this_net_peer
+        let mut gateway_ud = GatewayUserData::new();
```
neonphog: I don't think this is an invalid pattern, but it does add some readability cognitive overhead when compared to just being able to assume the "user_data" is equivalent to a mutable reference to the owning instance.

It's also a little complicated to pass in a mutable self reference while you're still in the constructor : ), but I think, for other reasons, that we're going to want to come up with an initialization pattern in our protocols whereby our constructors do no work, and we have "Initialize" RequestToActor and RequestToActorResponse types that we wait for before doing further processing.

(this can be a follow-on task)

crates/lib3h/src/engine/real_engine.rs
```diff
@@ -155,20 +193,91 @@ impl<'engine> RealEngine<'engine> {
...
+            self.space_gateway_map
+                .get_mut(&chain_id)
+                .expect("Should have the space gateway")
+        } else {
+            &mut self.network_gateway
+        };
+        gateway
+            .request(
+                Lib3hTrace,
+                GatewayRequestToChild::Dht(DhtRequestToChild::RequestThisPeer),
```
neonphog: The result of this call is precisely what we put in when creating this gateway, no?

If we want to keep our actor model pure and not make concrete state calls on the gateway actors directly, I'd say we should wrap the references with an internal struct here that can keep additional information like this so we don't have to make an async request to get it.

Alternately, if we want to side-step the actor model a bit in the short-term, I'd say making a direct request on the concrete gateway would be better than making a manual process call here in this helper function.

crates/lib3h/src/engine/real_engine.rs
```diff
@@ -155,20 +193,91 @@ impl<'engine> RealEngine<'engine> {
...
+                    } else {
+                        panic!("bad response to bind: {:?}", response);
+                    }
+                    Ok(())
+                }),
+            )
+            .expect("sync functions should work");
+        self.network_gateway
+            .process(&mut self.gateway_user_data)
+            .unwrap(); // FIXME unwrap
```
neonphog: Unlike my previous comment, this one actually needs to be realtime. Ideally we'd make the functions that need this list just actually wait on the async response (i.e. the code that follows this would move inside the callback.)

If we want to side-step that to get something working faster, I'd recommend switching this to a direct function call on the concrete gateway.

crates/lib3h/src/engine/real_engine.rs
```diff
@@ -285,9 +394,19 @@ impl<'engine> RealEngine<'engine> {
...
-                self.network_gateway.as_transport_mut().post(cmd)?;
+                // no-op ?
+                // send empty message so it connects
+                let cmd = GatewayRequestToChild::Transport(
+                    transport::protocol::RequestToChild::SendMessage {
+                        uri: msg.peer_uri,
+                        payload: Opaque::new(),
+                    },
+                );
+                self.network_gateway.publish(cmd)?;
```
neonphog: Shouldn't this be a request instead of a publish, so we can respond with a ConnectResult with the matching request_id from core?

crates/lib3h/src/engine/real_engine.rs
```diff
@@ -646,19 +776,22 @@ impl<'engine> RealEngine<'engine> {
             space_address,
             peer.peer_address,
         );
-        self.network_gateway
-            .as_transport_mut()
-            .send_all(&payload)
-            .ok();
+        let _res = self
+            .network_gateway
+            .publish(GatewayRequestToChild::SendAll(payload));
```
neonphog: `?`

crates/lib3h/src/engine/real_engine.rs
```diff
@@ -101,48 +99,88 @@ impl<'engine> RealEngine<'engine> {
...
+                NETWORK_GATEWAY_ID,
+                memory_network_endpoint,
+                dht_factory,
+                &dht_config,
+            )),
+            "network_gateway_",
         );
+        debug!("New MOCK RealEngine {} -> {:?}", name, this_net_peer);
         let transport_keys = TransportKeys::new(crypto.as_crypto_system())?;
+        let multiplexer = TransportMultiplex::new(Box::new(memory_transport));
```
neonphog: multiplexer is designed to take the network_gateway as its inner_transport.

... I'm not sure how exactly to go about that... do we need to implement a de-mangler that sits on top of the gateway that only sends / receives the transport bits of the protocol?

crates/lib3h/src/gateway/gateway_transport.rs
```diff
@@ -192,65 +63,180 @@ impl<'gateway> P2pGateway<'gateway> {
...
+                }
+                // Done
+                Ok(())
+            }),
+        )
+    }
+}
+
+/// Private internals
+impl P2pGateway {
```
neonphog: this impl ended up identical to the one above, should we merge them?

crates/lib3h/src/gateway/gateway_transport.rs
```diff
@@ -192,65 +63,180 @@ impl<'gateway> P2pGateway<'gateway> {
...
+                        )))?;
+                        Ok(())
+                    }),
+                );
+            }
+            transport::protocol::RequestToChild::SendMessage { uri, payload } => {
+                // uri is actually a dht peerAddress
+                // get actual uri from the inner dht before sending
+                let dht_uri_list = self.address_to_uri(&[&uri.to_string()])?;
+                let dht_uri = &dht_uri_list[0];
```
neonphog: I think we should make this do an actual async request on the dht and send the data in the callback.

We can also stop doing the send to multiple... we decided the added code complexity didn't warrant the potential future network throughput benefits

crates/lib3h/src/transport/memory_mock/ghost_transport_memory.rs
```diff
@@ -7,7 +11,7 @@ use url::Url;
 #[derive(Debug)]
 #[allow(dead_code)]
 enum RequestToParentContext {
-    Source { address: Url },
+    Source { uri: Url },
```
neonphog: :+1:

crates/lib3h/src/transport/protocol.rs
```diff
@@ -1,66 +1,44 @@
...
-    SendMessage,
+    SendMessage { payload: Opaque },
 }
 
 #[derive(Debug, Clone)]
 pub enum RequestToParent {
-    IncomingConnection { address: Url },
-    ReceivedData { address: Url, payload: Opaque },
-    TransportError { error: TransportError },
+    ErrorOccured { uri: Url, error: TransportError },
```
neonphog: We shouldn't need the uri here. If you get an error during a `SendMessage` for example, you will send an `Err` result back with the given request_id. If a connection is dropped for any reason between calls, I think it can be ignored. When you again try to send data on that connection, it should automatically re-connect or send back an error, again on the `SendMessage` response.

This error without the uri was intended for things like the bound network interface gets disconnected. (Hot swap-able NIC?)

crates/lib3h/src/engine/real_engine.rs
zippy: I will work on this whole initialization thing inside ghost_engine.  I think we can have a single constructor that takes the multiplexer, and then a bootstrap functions.

zippy: I've rebased ghost-engine off of this branch and can make progress there...

crates/lib3h/src/engine/real_engine.rs
ddd-mtl: not allways. In one occasion I need the up-to-date timestamp to know if I need to broadcast myself

PR #320, opened 2019-09-11
GhostActor tracing

maackle:
## PR summary

AKA `span`s everywhere!

Threads tracing spans through various function calls and channel messages to enable tracing across the network and between GhostActors. Uses child_of and follower_of semantics to relate these spans in a hierarchy.

## Followups

- [ ] address all `todo` and `fixme` spans
- [ ] review all child/follow decisions
- [ ] name everything appropriately
- [ ] address review comments

## Review checklist 


crates/lib3h/src/engine/network_layer.rs
```diff
@@ -22,13 +22,14 @@ impl<'engine> GhostEngine<'engine> {
         // Process the network gateway
         detach_run!(&mut self.multiplexer, |ng| ng.process(self))?;
         for mut request in self.multiplexer.drain_messages() {
+            let span = request.span().child("process_multiplexer");
```
zippy:
```suggestion
            // this should probably be changed so each arm has a different name span
            let span = request.span().child("process_multiplexer");
```

crates/lib3h/src/transport/transport_encoding/mod.rs
```diff
@@ -101,19 +101,24 @@ impl TransportEncoding {
             TransportError,
         >,
     ) -> TransportResult<()> {
+        let span = msg.span().child("handle_msg_from_inner");
```
zippy:
```suggestion
        // different name span for each arm?
        let span = msg.span().child("handle_msg_from_inner");
```

PR #336, opened 2019-09-13
Ghost Websocket Transport

### LINKED ISSUE #318: GhostActorify WS Transport
Examine how in-mem transport is done, do similar with WS transport.

lucksus:
## PR summary
closes #318 

Reuses some of the old `TransportWss` code and wraps it in an implementation of `GhostActor`.
The remainder of `TransportWss` lives on in `StreamManager` (in `streams.rs`).

## Review checklist 


crates/lib3h/src/transport/websocket/mod.rs
```diff
@@ -17,6 +16,39 @@ use std::{
...
+/// Commands that can be sent to an implementor of the Transport trait and handled during `process()`
+#[derive(Debug, PartialEq, Clone)]
+pub enum TransportCommand {
+    Connect(Url, String),
+    Send(Vec<ConnectionId>, Vec<u8>),
+    SendAll(Vec<u8>),
+    Close(ConnectionId),
+    CloseAll,
+    Bind(Url),
+}
```
lucksus: I've pulled these two enums back in again to get the old websocket code working, but I believe this one (`TransportCommand`) can go again easily.

crates/lib3h/src/lib.rs
```diff
@@ -4,6 +4,7 @@ extern crate detach;
 extern crate hcid;
 extern crate lib3h_crypto_api;
 extern crate lib3h_protocol;
+//#[macro_use]
```
neonphog:
```suggestion
```

crates/lib3h/src/transport/websocket/streams.rs
```diff
@@ -1,71 +1,33 @@
...
-type SocketMap<T> = std::collections::HashMap<String, WssInfo<T>>;
+/// how often should we send a heartbeat if we have not received msgs
+pub const DEFAULT_HEARTBEAT_MS: usize = 2000;
+
+/// when should we close a connection due to not receiving remote msgs
+pub const DEFAULT_HEARTBEAT_WAIT_MS: usize = 5000;
+
+/// a connection identifier
+pub type ConnectionId = String;
+pub type ConnectionIdRef = str;
```
neonphog: Do we need these? we deleted connection-id from the new protocol

crates/lib3h/src/transport/websocket/actor.rs
```diff
@@ -0,0 +1,442 @@
...
+
+                wait_for_message!(
+                    vec![&mut transport1, &mut transport2],
+                    t2_endpoint,
+                    "ReceivedData \\{ uri: \"wss://127\\.0\\.0\\.1:\\d+/\", payload: \"test message\" \\}"
+                );
+            }
+
+            trace!("Try {} successful!", index);
+            thread::sleep(time::Duration::from_millis(1000));
```
neonphog: Do we need this? It would be awesome to keep sleeps to less than 10 ms

crates/lib3h/src/transport/websocket/actor.rs
```diff
@@ -0,0 +1,442 @@
...
+                        .unwrap_or_else(|| {
+                            Url::from_str("wss://0.0.0.0")
+                                .expect("URL literal is syntactically correct")
+                        });
+                    self.endpoint_self.publish(
+                        Lib3hSpan::todo(),
+                        RequestToParent::ErrorOccured { uri, error },
+                    )?;
+                }
+                StreamEvent::ConnectResult(_connection_id, _) => {}
```
neonphog: When the user of this lib sends in a SendMessage, and we are not already connected to that remote URI, the ghost actor contract is for the SendMessageResult response to only be sent out after all the connection handshaking is complete and the actual low-level send is invoked.

This is an awesome iteration, I'm totally cool with putting a comment to this effect here and iterating as a follow-on.

crates/lib3h/src/transport/websocket/streams.rs
lucksus: Yes, `StreamManager` (former `TransportWss`) still uses connection IDs. I thought about completely getting rid of those, but then decided to leave the old code as is as much as possible - and just find a clean cut. Could also change the interface of `StreamManager` to only talk about URIs. Do you think it would be worth doing so now?

crates/lib3h/src/transport/websocket/actor.rs
```diff
@@ -0,0 +1,518 @@
...
+        match msg
+            .take_message()
+            .expect("GhostMessage must have inner RequestToChild")
+        {
+            RequestToChild::SendMessage { uri, payload } => {
+                match self.bound_url.clone() {
+                    None => {
+                        let _ = msg.respond(Err(TransportError::new(
+                            "Transport must be bound before sending".to_string(),
+                        )));
```
neonphog: `?`

crates/lib3h/src/transport/websocket/actor.rs
```diff
@@ -0,0 +1,518 @@
...
+                        // Send it data from us
+                        let _ = match self.streams.send(&uri, &bytes) {
+                            Ok(()) => msg.respond(Ok(RequestToChildResponse::SendMessageSuccess)),
+                            Err(error) => {
+                                println!("Error during send: {:?}", error);
+                                let payload = Opaque::from(bytes);
+                                msg.put_message(RequestToChild::SendMessage { uri, payload });
+                                return Err(msg);
+                            }
+                        };
```
neonphog: `?`

PR #338, opened 2019-09-13
Simchat GhostActor Part 2

willemolding:
## PR summary

- Refactors the simchat to multiple files and adds a trait for making alternative implementation (e.g. the holochain version) compatible with the same interface
- Refactor state to be a struct rather than in the scope
- Adds chat events for channel messages (receive, post, query) which call the respective lib3h messages
- Add channel messages on a single time anchor
- Add coloring to text output
- Adds managing of the author list and gossip list

This PR gets most of the way to working channel messages. The only limitation for now is that all messages are held on a single base. Once tested with a working networking implementation it can be easily updated to use proper time anchors. SImChat keeps track of which messages have been viewed already and so subsequent polls will not print duplicate messages.

SimChat nodes have their own local store which lib3h can both store and retrieve data from via the Lib3hToClient messages. Messages are stored as EntryAspects on a time anchor entry. At this stage I am not sure if these aspects will be correctly gossiped (e.g. can the GetEntryAuthorList return aspects or only entries?). This may need to be slightly modified (e.g. messages as actual entries) to work properly. The state keeps track of the AuthorList and GossipList. This may cause problems if leaving and rejoining the network as the store is in memory only.

## Review checklist 


crates/tools/sim_chat/src/lib3h_simchat/handle_lib3h_event.rs
```diff
@@ -0,0 +1,321 @@
...
+    protocol::{Lib3hToClient, Lib3hToClientResponse},
+};
+
+///
+pub fn handle_and_convert_lib3h_event(
+    lib3h_message: Lib3hToClient,
+    state: &mut Lib3hSimChatState,
+) -> (
+    Option<ChatEvent>,
+    Option<Result<Lib3hToClientResponse, Lib3hError>>,
```
zippy: This can be a `Lib3hResult<Lib3hToClientResponse>`

crates/tools/sim_chat/src/lib3h_simchat/mod.rs
```diff
@@ -0,0 +1,443 @@
...
+
+                    std::thread::sleep(std::time::Duration::from_millis(10));
+                }
+            })),
+            thread_continue,
+            out_send,
+        }
+    }
+
+    /// Uses n3h to try and bootstrap via the given URI
```
zippy:
```suggestion
    /// Uses lib3h to try and bootstrap via the given URI
```

crates/tools/sim_chat/src/lib3h_simchat/mod.rs
```diff
@@ -0,0 +1,443 @@
...
+    fn calling_join_space_triggers_success_message() {
+        let (s, r) = crossbeam_channel::unbounded();
+        let mut chat = new_sim_chat_mock_engine(Box::new(move |event| {
+            s.send(event.to_owned()).expect("send fail");
+        }));
+
+        chat.send(join_event());
+
+        let chat_messages = r.iter().take(2).collect::<Vec<_>>();
+        assert_eq!(chat_messages, vec![join_event(), join_success_event(),],);
```
willemolding: These tests should not depend on the ordering of the messages as this can be changed at any time

Issue #324, opened 2019-09-12
GhostEngine: handle the three sub-process calls


zippy:
process_concrete() needs to call the network/space/ugly process loops, but these all currently return legacy Lib3hServerProtocol commands which need to be handled in the callbacks of those processors instead.

PR #344, opened 2019-09-13
WIP - ghost engine process()

ddd-mtl:
and removed RealEngine

#324


crates/lib3h_protocol/src/protocol.rs
```diff
@@ -58,10 +58,14 @@ pub enum ClientToLib3hResponse {
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub enum Lib3hToClient {
     // -- Connection -- //
+    /// Notification of successful connection to a network
+    Connected(ConnectedData),
```
neonphog: This is a response to `ClientToLib3h::Connect`, we don't need to add it here, it's already `ClientToLib3hResponse::ConnectResult`

crates/lib3h_protocol/src/protocol.rs
```diff
@@ -58,10 +58,14 @@ pub enum ClientToLib3hResponse {
...
     /// Notification of disconnection from a network
     Disconnected(DisconnectedData),
 
     // -- Direct Messaging -- //
+    /// the response received from a previous `SendDirectMessage`
+    SendDirectMessageResult(DirectMessageData),
```
neonphog: This is a response to `ClientToLib3h::SendDirectMessage`, we don't need to add it here, it's already `ClientToLib3hResponse::SendDirectMessageResult`

crates/lib3h/src/engine/space_layer.rs
```diff
@@ -156,7 +148,9 @@ impl RealEngine {
                             provider_agent_id: chain_id.1.clone(),
                             aspect_address_list: None,
                         };
-                        outbox.push(Lib3hServerProtocol::HandleFetchEntry(msg_data))
+                        self.lib3h_endpoint
+                            .publish(Lib3hSpan::todo(), Lib3hToClient::HandleFetchEntry(msg_data))
```
zippy: I think this needs to be a request not a publish, right?

crates/lib3h_protocol/src/protocol.rs
ddd-mtl: Its not a response to a Connect since `Connect` is a request to connect to a specific peer.
`Connected` is a notification that we are connected to the **network**.

neonphog: ... as in the network health stuff that gives us partition warnings, etc? I was assuming we weren't implementing any of that, since we haven't fleshed out the details of the core/lib3h protocol for it.

ddd-mtl: Since, in the end, we are not going to have a "Connect" request in the first place I figured the only use for this would be for the network health thing yes.

crates/lib3h_protocol/src/protocol.rs
ddd-mtl: Its used for doing a publish() so it can't be in the Response enum.
We can't do a respond() to a ClientToLib3h::SendDirectMessage since there is no response at the transport protocol layer.

neonphog: Right! yes, I got this one backwards, sorry : )

crates/lib3h_protocol/src/protocol.rs
neonphog: Maybe we need to rename this. I think we're going to still need a "Bootstrap" message, i.e. if some guy is riding by with a wifi holochain node on his bicycle and a QR code to connect to it. Your ui will pass it in ultimately to lib3h and need to know that particular "Bootstrap" message was a success or not, so will need a "BootstrapResult".

PR #345, opened 2019-09-13
GhostActor TransportEncoding

neonphog:
## Review checklist 


crates/lib3h/src/transport/protocol.rs
```diff
@@ -50,6 +52,78 @@ pub type DynTransportActor = Box<
...
+                    Ok(())
+                }),
+            )?;
+        }
+        detach_run!(&mut self.endpoint_inner, |it| it.process(self))?;
+        for mut msg in self.endpoint_inner.as_mut().drain_messages() {
+            let data = msg.take_message().expect("exists");
+            self.endpoint_self.publish(Lib3hSpan::todo(), data)?;
+        }
+        Ok(false.into())
```
dymayday: So, Is this function always tells us that the work was never done ? :thinking:

neonphog: hehe, yeah, just a stub until @struktured gives us a pattern to follow : )

crates/lib3h/src/transport/protocol.rs
```diff
@@ -50,6 +52,78 @@ pub type DynTransportActor = Box<
     >,
 >;
 
+/// this is awkward... I *think* it'll go away with ghost-actor proc-macro
+pub struct TransportEndpointAsActor {
```
struktured: About to say it was awkward and hope you're right about the macro eliminating it.

PR #377, opened 2019-09-20
Implement drop for web socket transport

struktured:
## PR summary

This PR upgrades `tungstenite` and dependent crate `url` so we can gracefully send close commands to peers before dropping. The stream manager, before closing a socket, writes a close message to the other end. The web socket `Drop` trait is implemented and invokes `close_all` on the stream manager.

## Review checklist 
- [ ] The story has unit or integration tests _TODO!!!_

### DIFF
--- a/crates/lib3h/src/transport/websocket/streams.rs
+++ b/crates/lib3h/src/transport/websocket/streams.rs
@@ -389,6 +389,7 @@ impl<T: Read + Write + std::fmt::Debug> StreamManager<T> {
                         Err(e.into())
                     }
                     Err(tungstenite::error::Error::ConnectionClosed) => {
+                        error!("Connection unexpectedly closed");
                         // close event will be published
                         Ok(())
                     }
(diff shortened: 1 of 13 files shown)


crates/lib3h/src/transport/websocket/streams.rs
```diff
@@ -388,7 +388,7 @@ impl<T: Read + Write + std::fmt::Debug> StreamManager<T> {
                         }
                         Err(e.into())
                     }
-                    Err(tungstenite::error::Error::ConnectionClosed(_)) => {
+                    Err(tungstenite::error::Error::ConnectionClosed) => {
                         // close event will be published
                         Ok(())
```
dymayday: Should we have a log that we encountered an error here ?

crates/lib3h/src/transport/websocket/streams.rs
```diff
@@ -420,7 +420,7 @@ impl<T: Read + Write + std::fmt::Debug> StreamManager<T> {
...
```
dymayday: Same

crates/lib3h/src/transport/websocket/streams.rs
struktured: Done

PR #379, opened 2019-09-20
Lib3hUri + peerData Renames + timestamp fix

ddd-mtl:
Changed all use of `url::Url` to our own `Lib3hUri` that can convert to `url::Url`.

Did the following renames:
`peer_address` -> `peer_name`
`peer_uri` -> `peer_location`
`machine_id` -> `transport_id`

Now generates timestamp on peerData creation.
Upped default timeout so tests can pass.


crates/lib3h/src/transport/memory_mock/ghost_transport_memory.rs
```diff
@@ -48,14 +47,14 @@ pub type GhostTransportMemoryEndpointContextParent = GhostContextEndpoint<
...
 #[allow(dead_code)]
 pub struct GhostTransportMemory {
-    machine_id: Address,
+    transport_id: Address,
     network: Arc<Mutex<MemoryNet>>,
     endpoint_parent: Option<GhostTransportMemoryEndpoint>,
     endpoint_self: Detach<GhostTransportMemoryEndpointContext>,
     /// My peer uri on the network layer (not None after a bind)
-    maybe_my_address: Option<Url>,
+    maybe_my_address: Option<Lib3hUri>,
```
struktured: Is this the same as `Lib3hUrl::undefined("")` now, or not?

crates/lib3h_protocol/src/uri.rs
```diff
@@ -0,0 +1,188 @@
...
+    pub fn with_undefined(other: &str) -> Self {
+        let url = Url::parse(&format!("{}:{}", UNDEFINED_SCHEME, other)).unwrap();
+        Lib3hUri(url)
+    }
+    pub fn with_memory(other: &str) -> Self {
+        let url = Url::parse(&format!("{}://{}", MEMORY_SCHEME, other)).unwrap();
+        Lib3hUri(url)
+    }
+}
+
```
struktured: I'd recommend making a private function that looks something like
```Rust
fn parse_internal(url: &str) -> Url {
 Url::parse(str).expect(format!("Invalid url: {:?}", str))
}
and replacing all the calls parse above with that function instead. Reason being I find it frustrating that the `unwrap()` version doesn't tell which string failed to parse.
```

crates/zombie_actor/src/ghost_tracker.rs
```diff
@@ -3,7 +3,7 @@ use std::collections::HashMap;
 
 use crate::{ghost_error::ErrorKind, GhostError, GhostResult, RequestId, WorkWasDone};
 
-const DEFAULT_TIMEOUT: std::time::Duration = std::time::Duration::from_millis(2000);
+const DEFAULT_TIMEOUT: std::time::Duration = std::time::Duration::from_millis(20000);
```
struktured: Ouch, 20 seconds a long time to make sure a test passes- do you know which one was sensitive to this timing?

crates/lib3h_protocol/src/uri.rs
```diff
@@ -0,0 +1,188 @@
...
+            TRANSPORT_SCHEME, transport_id, agent_id
+        ))
+        .unwrap();
+        Lib3hUri(url)
+    }
+    pub fn with_agent_id(agent_id: &Address) -> Self {
+        let url = Url::parse(&format!("{}:{}", AGENT_SCHEME, agent_id)).unwrap();
+        Lib3hUri(url)
+    }
+    pub fn with_undefined(other: &str) -> Self {
```
zippy: What I had intended with this was to create a "NULL" URI, i.e. what that was literally like `None`  which has the semantic meaning of we don't have a URI here, not just one where the scheme is the undefined scheme.  

I see you made the "Fixme's" into Undefined's and I don't think that's right.  They should be FIXME's within in one of the scheme's that they are supposed to be.

crates/lib3h/src/dht/mod.rs
```diff
@@ -57,8 +59,8 @@ pub mod tests {
         pub fn new() -> Self {
             DhtData {
                 this_peer: PeerData {
-                    peer_address: "FIXME".to_string(),
-                    peer_uri: Url::parse("fixme://host:123").unwrap(),
+                    peer_name: Lib3hUri::with_undefined("FIXME"),
+                    peer_location: Lib3hUri::with_undefined("host:123"),
```
zippy: See comment below on "undefined" URI type

crates/lib3h/src/dht/rrdht.rs
```diff
@@ -20,8 +19,8 @@ impl RrDht {
         RrDht {
             inbox: VecDeque::new(),
             this_peer: PeerData {
-                peer_address: "FIXME".to_string(),
-                peer_uri: Url::parse("fixme://host:123").unwrap(),
+                peer_name: Lib3hUri::with_undefined("FIXME"),
```
zippy: ditto, shouldn't be undefined.

crates/zombie_actor/src/ghost_tracker.rs
ddd-mtl: Added TODO

crates/lib3h/src/transport/memory_mock/ghost_transport_memory.rs
ddd-mtl: No it doesn't replace the option as it is about being bound or not.

crates/lib3h/src/engine/mod.rs
```diff
@@ -86,15 +85,15 @@ impl TransportKeys {
         let mut secret_key = crypto.buf_new_secure(crypto.sign_secret_key_bytes());
         crypto.sign_keypair(&mut public_key, &mut secret_key)?;
         Ok(Self {
-            transport_id: hcm0.encode(&public_key)?,
+            transport_id: hcm0.encode(&public_key)?.into(),
```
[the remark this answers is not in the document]
ddd-mtl: Here we are just converting a String into a HashString.

crates/lib3h/src/dht/rrdht.rs
ddd-mtl: We shouldn't allow `PeerData::new()` The only reason we have it is for unit-testing.

ddd-mtl:
@zippy There is no such thing as using `mem://addr_1..` or `wss://...` as transport_id.
transport_id is a HCID. Now we have a uri scheme for it just to make it convenient to store it as an Url, since this is the required format for the DHT peerData.
If you find transport_id's as mem://addr_1.. or wss://... URI's then something is wrong in the code.
Let me know if the latest changes resolves all the issues you raised.

crates/lib3h/src/transport/memory_mock/ghost_transport_memory.rs
```diff
@@ -411,30 +410,30 @@ mod tests {
             Some(bound_transport3_address.clone())
         );
 
-        // check that machine_ids were advertised
+        // check that transport_ids were advertised
```
zippy: Here's a place where we use the term transport_id and the code is looking for a `mem://`  Maybe the comment is the error?

PR #381, opened 2019-09-20
Discovery wss mdns

dymayday:
## PR summary

Use mDNS for bootstrapping nodes using websocket transport.

## Review checklist 


crates/lib3h/src/transport/websocket/actor.rs
```diff
@@ -8,35 +8,69 @@ use crate::transport::{
...
+// Use mDNS for bootstrapping
+use lib3h_mdns::{MulticastDns, MulticastDnsBuilder};
+
 pub type Message =
     GhostMessage<RequestToChild, RequestToParent, RequestToChildResponse, TransportError>;
 
 pub struct GhostTransportWebsocket {
     #[allow(dead_code)]
     machine_id: Address,
+    networkid_address: Address,
```
struktured: Use snake casing instead- `network_id_address`.

crates/lib3h/src/transport/websocket/actor.rs
```diff
@@ -548,4 +608,114 @@ mod tests {
     fn should_invoke_drop() {
         // TODO
     }
+
+    /// Check if we manage to discover nodes using WebSocket for bootstapping using mDNS.
+    #[test]
+    fn mdns_wss_bootstrapping_test() {
+        let networkid_address: Address = "wss-bootstapping-network-id.holo.host".into();
```
struktured: Use `nanoid`?

crates/lib3h/src/transport/websocket/actor.rs
dymayday: :+1:

crates/lib3h/src/transport/websocket/actor.rs
dymayday: :+1:

crates/lib3h/src/transport/websocket/actor.rs
```diff
@@ -8,35 +8,72 @@ use crate::transport::{
...
     pending: Vec<Message>,
+    mdns: Option<MulticastDns>,
 }
 
+// Here we just need to use mDNS, but use it only once, with advertise probably, and that's all.
 impl Discovery for GhostTransportWebsocket {
     fn advertise(&mut self) -> DiscoveryResult<()> {
-        Ok(())
+        // Lazily instantiate mDNS
+        if self.mdns.is_none() {
```
zippy: Do we want need to have an instance of the MDNS per memory transport, or can they all access one instance that lives in a mutex in a lazy static?

crates/lib3h/src/transport/websocket/actor.rs
```diff
@@ -132,6 +175,8 @@ impl GhostTransportWebsocket {
 
                     if let Ok(url) = maybe_bound_url {
                         self.bound_url = Some(url.clone());
+                        self.advertise()
```
zippy: We need to change the test so that it shows that this advertise worked after a bind.

zippy: done

crates/lib3h/src/transport/websocket/actor.rs
neonphog: We could do some kind of "discovery manager" concept. Some of the transports can share discovery mechanisms - i.e. native and websocket can both use mdns, while memory and something like bluetooth would probably need specialized mechanisms...

But we can also go with one discovery per transport, because I also think it's totally valid for say native and wss to each have their own mdns queries / responses.

dymayday: I agree with @neonphog , I think it's valid and it can avoid confusion as one shared mDNS instance in a lazy_static would return every Url for every instance of lib3h, instead of doing it only for its own network ID.

PR #395, opened 2019-09-24
timeout handling + ci fixes

neonphog:
## PR summary

https://app.zenhub.com/workspaces/networking-5d4050f96d87a44faafbe34d/issues/holochain/lib3h/388

^^ does NOT complete this. After two painful 24 minute long mac builds... they both pass.

Hopefully if we merge this, when we *do* get a timeout on macos, we'll at least know which one it was.

## Review checklist 


crates/zombie_actor/src/ghost_tracker.rs
```diff
@@ -254,7 +256,7 @@ mod tests {
                 // when the timeout happens the callback should get
                 // the timeout enum in the callback_data
                 match callback_data {
-                    GhostCallbackData::Timeout => me.state = "timed_out".into(),
+                    GhostCallbackData::Timeout(_) => me.state = "timed_out".into(),
```
struktured: Is this '_' useful data to record somewhere?

crates/zombie_actor/src/lib.rs
```diff
@@ -357,7 +357,7 @@ mod tests {
                             Box::new(move |_m:&mut GatewayTransport, response| {
 
                                 // got a timeout error
-                                if let GhostCallbackData::Timeout = response {
+                                if let GhostCallbackData::Timeout(_) = response {
```
struktured: Same comment as above.

crates/zombie_actor/src/ghost_tracker.rs
neonphog: Yep! I bubbled this up everywhere, except in a couple unit tests like this to keep them simple.

PR #397, opened 2019-09-24
Fix send message response test

struktured:
## PR summary

This PR uncomments the ignored `two_memory_nodes_basic_suite` and gets `test_send_message` to pass (the remaining tests in the suite are still commented out however). The critical fix, besides cleaning up the test itself with the `assert_msg_*` macros, is to map the replaced request_id back to the original request, especially for `HandleDirectMessageResult`.

Currently a hack is in place that maps them directly in ghost engine wrapper and only does the reversing magic if its a `HandleDirectMessageResult`. This might need to be generalized for other cases in the future.

## Review checklist 


crates/lib3h/src/engine/ghost_engine_wrapper.rs
```diff
@@ -281,22 +301,32 @@ where
...
+            Lib3hServerProtocol::FetchEntryResult(data) => data.request_id = request_id,
+            Lib3hServerProtocol::HandleFetchEntry(data) => data.request_id = request_id,
+            Lib3hServerProtocol::HandleStoreEntryAspect(data) => data.request_id = request_id,
+            Lib3hServerProtocol::HandleDropEntry(data) => data.request_id = request_id,
+            Lib3hServerProtocol::HandleQueryEntry(data) => data.request_id = request_id,
+            Lib3hServerProtocol::HandleGetAuthoringEntryList(data) => data.request_id = request_id,
+            Lib3hServerProtocol::HandleGetGossipingEntryList(data) => data.request_id = request_id,
+            Lib3hServerProtocol::HandleSendDirectMessage(data) => {
+                // Cache this request_id as the ghost engine layer needs it to associate
+                // the response message
```
neonphog: This is the part where I'm pretty sure something lower is relying on a request_id that shouldn't be... The lower level received a request_id over the P2p message, and should be capturing it in the callback when it makes the request up to here, and shouldn't be paying attention to the one we re-inject from the request_id_map above.

I'm totally cool with merging this, especially if it means we have a working test, so we can muck with the lower code and prove it still works :+1:

struktured: Agreed, would like to get rid of this request_id dependency!

PR #401, opened 2019-09-25
Send message failure passes

struktured:
## PR summary

This PR adds an attempt count to the send message protocol struct. It is used to stop attempting retries in the gateway transport after a constant number of times (= `MAX_RETRY_ATTEMPTS` which is currently set to 5).

The send message failure test is uncommented out and has been modernized. It passes after the aforementioned changes.

## Review checklist 


struktured:
> nice cleanup and fixes. We may find at some point we need to timeout on time instead of attempts, but we don't know that yet, and this works for now

Yeah I was thinking the same thing- if it does come up we can just change `attempt` to `timestamp` and adjust the logic in a couple places. The change would be quite minimal.

Issue #399, opened 2019-09-25
discovery integration test


neonphog:
make sure discovery / bootstrap connect works for both memory transport and wss transport

PR #404, opened 2019-09-26
Discovery in-memory / WSS tests

dymayday:
## PR summary

This PR proposes to add tests in order to make sure discovery / bootstrap connect works for both memory transport and wss transport.

Reference https://github.com/holochain/lib3h/issues/399

This work is irrelevant now because its fated to be discarded when ghost_actor v2 will kick off.

## Review checklist 


zippy:
I remembered what I did for discovery in the memory transport.  Look at try_discover()  This gets called in process_concrete() of GhostMemoryTransport as periodic discovery.  See line 183 of ghost_transport_memory.   So we need to do something similar for Websockets.
In the case of the in-memory transport we do this until some other node is found as bootstrapping, assuming that all further nodes will be found through gossip.  The same assumption is reasonable for websockets.
Thus @meidhy you only need to add something similar for websockets, not any discover call in the ghost engine constructor like I said earlier.

crates/lib3h/src/transport/websocket/actor.rs
```diff
@@ -338,6 +345,64 @@ impl GhostTransportWebsocket {
...
     }
+
+    /// Try to discover peers on the network using mDNS.
+    fn try_discover(&mut self) {
+        self.last_discover = match self.last_discover {
+            None => Some(Instant::now()),
+            Some(last_discover) => {
+                // Let's check if it time to discover some peers, but only if we already
+                // did some url binding
+                if self.bound_url.is_none()
```
zippy: Doesn't this have to be `.is_some()`?

dymayday: Oups, good catch !!

dymayday:
This work is irrelevant now because its fated to be discarded when ghost_actor v2 will kick off.

So we can merge it as soon as the CI pass IMO.

PR #406, opened 2019-09-26
Do not capture backtraces by default

neonphog:
## PR summary

- adds some benchmarking code that *can* run in CI, but doesn't by default. You can also run `cargo bench` or `MAKE_ENV=local make benchmarks`
- causes backtrace logic to NOT capture by default. It turns out macOs and Windows cannot handle it

## Review checklist 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -14,9 +14,7 @@ jobs:
    working_directory: ~/repo
    steps:
      - checkout
-     - run:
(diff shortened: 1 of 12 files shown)


zippy:
But it's sad that now there's another MacOS CI problem:: https://circleci.com/gh/holochain/lib3h/2830?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Issue #411, opened 2019-09-27
FIx author aspect test


struktured:
In `two_basic.rs`,  `test_author_one_aspect` fails. Fix it.

ddd-mtl:
https://github.com/holochain/lib3h/pull/419

PR #419, opened 2019-09-27
Fixed authoring test

ddd-mtl:
#411
Might not be the best solution
Also dont know what the latest workflow is now when asking for an non-existing entry

### DIFF
--- a/crates/lib3h/src/engine/ghost_engine.rs
+++ b/crates/lib3h/src/engine/ghost_engine.rs
@@ -754,8 +754,8 @@ fn includes(list_a: &[Address], list_b: &[Address]) -> bool {
     let set_b: HashSet<_> = list_b.iter().map(|addr| addr).collect();
     set_b.is_subset(&set_a)
 }
-
-pub fn handle_gossip_to<
+#[allow(non_snake_case)]
+pub fn handle_GossipTo<
     G: GhostActor<
         GatewayRequestToParent,
         GatewayRequestToParentResponse,
@@ -769,7 +769,7 @@ pub fn handle_gossip_to<
     gossip_data: GossipToData,
 ) -> Lib3hResult<()> {
     debug!(
-        "({}) handle_gossip_to: {:?}",
+        "({}) handle_GossipTo: {:?}",
         gateway_identifier, gossip_data,
     );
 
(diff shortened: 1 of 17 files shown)


crates/lib3h/src/engine/ghost_engine.rs
```diff
@@ -782,11 +782,11 @@ pub fn handle_gossip_to<
         //            }
         //            // TODO END
 
-        // Convert DHT Gossip to P2P Gossip
+        // Convert DHT *GossipTo* to P2P *Gossip*
         let p2p_gossip = P2pProtocol::Gossip(GossipData {
             space_address: gateway_identifier.clone(),
             to_peer_name: to_peer_name.clone(),
-            from_peer_name: Lib3hUri::with_undefined(), // FIXME
+            from_peer_name: Lib3hUri::with_undefined(), // Should be gateway.as_mut().as_mut().this_peer().peer_location,
```
struktured: Is this breaking any functionality or is from_peer_name not a super important field?

crates/lib3h/src/engine/ghost_engine_wrapper.rs
```diff
@@ -266,7 +271,7 @@ where
     /// Process Lib3hClientProtocol message inbox and
     /// output a list of Lib3hServerProtocol messages for Core to handle
     pub fn process(&mut self) -> Lib3hProtocolResult<(DidWork, Vec<Lib3hServerProtocol>)> {
-        trace!("[legacy engine] process");
+        // trace!("[legacy engine] process");
```
struktured:
```suggestion
```

crates/lib3h/src/engine/space_layer.rs
```diff
@@ -325,7 +325,21 @@ impl<'engine> GhostEngine<'engine> {
...
+                let space_gateway =
+                    self.get_space(&gossip_data.space_address, &gossip_data.to_peer_name.into())?;
+                space_gateway.publish(
+                    span.follower("TODO"),
+                    GatewayRequestToChild::Dht(DhtRequestToChild::HandleGossip(remote_gossip)),
+                )?;
+            }
+            _ => {
+                panic!("can't handle space layer receive of {:?}", p2p_msg);
+            }
```
struktured: Was this the critical missing piece for the test to pass?

ddd-mtl: A half of it yes. The other half is having gateway bubble up the dht request to engine.

crates/lib3h/src/engine/ghost_engine.rs
ddd-mtl: Not yet it seems but it should be a breaking functionality otherwise the field is useless and should be removed.

Issue #414, opened 2019-09-27
linux ci flaky failure


neonphog:
https://circleci.com/gh/holochain/lib3h/2849?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

```
---- transport::websocket::actor::tests::test_websocket_transport stdout ----
thread 'transport::websocket::actor::tests::test_websocket_transport' panicked at 'called `Result::unwrap()` on an `Err` value: GhostError(Other("TransportError: Unknown error encountered: \'Operation not permitted (os error 1)\'."))', src/libcore/result.rs:1051:5
stack backtrace:
  <..snip..>
  12: lib3h::transport::websocket::actor::tests::test_websocket_transport
             at crates/lib3h/src/transport/websocket/actor.rs:515
```

struktured:
Found another error using the flaky test finder:

```Rust
thread 'transport::websocket::actor::tests::test_websocket_transport_reconnect' panicked at 'assertion failed: `(left == right)`
  left: `None`,
 right: `Some(Lib3hUri("wss://127.0.0.1:2026/"))`', crates/lib3h/src/transport/websocket/actor.rs:632:17
stack backtrace:
test transport::websocket::actor::tests::test_websocket_transport ... ok
Try 1 successful!
Try 2 successful!
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.29/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.29/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:47
   3: std::sys_common::backtrace::print
             at src/libstd/sys_common/backtrace.rs:36
   4: std::panicking::default_hook::{{closure}}
             at src/libstd/panicking.rs:200
   5: std::panicking::default_hook
             at src/libstd/panicking.rs:214
   6: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:477
   7: std::panicking::continue_panic_fmt
             at src/libstd/panicking.rs:384
   8: std::panicking::begin_panic_fmt
             at src/libstd/panicking.rs:339
   9: lib3h::transport::websocket::actor::tests::test_websocket_transport_reconnect
[...]

PR #430, opened 2019-09-30
Fix test websocket transport on ci

struktured:
## PR summary

This PR extends the `ghost_test_harness` to include a macro will retry a test with different inputs if it fails until the test finally passes. After a fixed number of unsuccessful iterations it will fail regardless.

This works contributes to issue #414 but will not 100% solve it as several other places have the same binding problem. A subsequent PR will reuse this code in other critical places.

## Review checklist 


crates/zombie_actor/src/ghost_test_harness.rs
```diff
@@ -290,18 +298,57 @@ macro_rules! wait1_for_callback {
...
+                assert_eq!($re, actual.as_str());
+            }
+        }
+        is_match
+    }};
+}
+
+#[allow(unused_macros)]
+#[macro_export]
+macro_rules! wait1_for_repeatable_callback {
```
zippy: It took me quite a bit headbanging to understand how to use this function.  It's very cool, but I would have been helped by a comment above the macro definition explaining how build the request_fn and how it's return type works for passing the state along.  Alternatively the comment could go in the test.  I ended up having to triangulate the test (which is just a binary state) and the usage for websockets, which keeps changing the state.

crates/lib3h/src/transport/websocket/actor.rs
```diff
@@ -462,32 +463,38 @@ mod tests {
...
-                    );
-                    Ok(())
-                }),
-            )
-            .unwrap();
+        let init_transport1_address: Lib3hUri = Url::parse(&format!("wss://127.0.0.1:{}", 1024))
+            .unwrap()
+            .into();
+
+        let request_fn = Box::new(|transport1_address: Lib3hUri| {
```
zippy:
```suggestion
        // closure for the wait1_for_repeatable_callback! that builds a new request with a different port number
        // an a modified regular expression expectation to match
        let request_fn = Box::new(|transport1_address: Lib3hUri| {
```

crates/zombie_actor/src/ghost_test_harness.rs
struktured: I totally understand that. I documented the macro and will provide simpler versions that build on top of it in subsequent PRs.

Issue #211, opened 2019-07-17
Time Mock


ddd-mtl:
Implement a TimeMock and have `since_epoch_ms()` us it in tests in order to get deterministic time, as opposed to the flaky SystemTime currently used.

AC:
 - In tests, timestamp values are consistent between runs.

PR #436, opened 2019-09-30
add ping pong message types to p2p protocol

neonphog:
## PR summary

- Makes a start at using CapnProto for these new message types - doubly wrapped by the message-pack p2p-protocol for now to make iterative transition possible. i.e. we can move messages over one at a time. When we get them all moved, we can drop the messagepack layer.
- Switches to using Ping / Pong instead of empty (zero byte) messages to establish connections.
- Created a follow-on task to integrate the ping/pong more tightly with our lib3h/core protocol flows - https://app.zenhub.com/workspaces/networking-5d4050f96d87a44faafbe34d/issues/holochain/lib3h/446

## Review checklist 


crates/lib3h/src/gateway/gateway_dht.rs
```diff
@@ -75,13 +76,17 @@ impl P2pGateway {
                     self.identifier.nickname, peer_data.peer_name, peer_data.peer_location,
                 );
                 // Send phony SendMessage request so we connect to it
```
zippy:
```suggestion
                // Send Ping so we connect to it
```

crates/lib3h/src/gateway/gateway_transport_send.rs
```diff
@@ -176,6 +176,21 @@ impl P2pGateway {
...
         cb: SendCallback,
     ) -> GhostResult<()> {
+        if !self.message_encoding.is_attached() {
+            // we must be in a call chain resulting from an incoming message
+            // we need to wait for the next process() call to continue
+            return self.priv_send_queue_pending(SendMetaData {
+                send_data: SendData::WithFullLowUri(send_data),
+                // we want to invoke this on the very next process call
+                // set our last_attempt back far enough to ensure this
+                last_attempt: std::time::Instant::now()
```
zippy: Should we be able to pass in None to `priv_send_queue_pending` so that it can do this if we want a "send immediate
"?

crates/p2p_protocol/src/p2p.rs
```diff
@@ -0,0 +1,163 @@
...
+    MsgPong(MsgPong),
+}
+
+/// get the current system milliseconds since unix epoch
+fn now_ms() -> u64 {
+    let out = std::time::SystemTime::now()
+        .duration_since(std::time::UNIX_EPOCH)
+        .expect("duration since unix epoch");
+    out.as_secs() * 1000 + u64::from(out.subsec_nanos()) / 1_000_000
+}
```
ddd-mtl: A bit worried of not using our `since_epoch_ms()` here, i.e. of not using a centralized clock.

crates/p2p_protocol/src/p2p.rs
```diff
@@ -0,0 +1,163 @@
...
+    /// create a new ping message
+    /// if `send_epoch_ms` is None, will be set to now
+    pub fn create_ping(mut send_epoch_ms: Option<u64>) -> Self {
+        if send_epoch_ms.is_none() {
+            send_epoch_ms = Some(now_ms());
+        }
+        P2pMessage::MsgPing(MsgPing {
+            ping_send_epoch_ms: send_epoch_ms.unwrap(),
+        })
+    }
```
ddd-mtl:
```
    pub fn create_ping(mut send_epoch_ms: Option<u64>) -> Self {
        P2pMessage::MsgPing(MsgPing {
            ping_send_epoch_ms: send_epoch_ms.unwrap_or(now_ms()),
        })
    }
```

crates/p2p_protocol/src/p2p.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    /// create a new pong message
+    /// if `recv_epoch_ms` is None, will be set to now
+    pub fn create_pong(send_epoch_ms: u64, mut recv_epoch_ms: Option<u64>) -> Self {
+        if recv_epoch_ms.is_none() {
+            recv_epoch_ms = Some(now_ms());
+        }
+        P2pMessage::MsgPong(MsgPong {
+            ping_send_epoch_ms: send_epoch_ms,
+            ping_received_epoch_ms: recv_epoch_ms.unwrap(),
```
ddd-mtl: same here

crates/p2p_protocol/src/p2p.rs
```diff
@@ -0,0 +1,163 @@
+use crate::{error::P2pResult, p2p_capnp};
+
+/// a message used to verify connectivity with a remote node
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq, Hash)]
+pub struct MsgPing {
+    /// set the milliseconds since unix epoch when sending this message
+    pub ping_send_epoch_ms: u64,
```
ddd-mtl: `ping_` part of the variable name seems unnecessary

crates/p2p_protocol/src/p2p.rs
neonphog: agree, but, alas, they are in separate crates... Added a note to #211 -

> Also we're using SystemTime in at least two different crates (lib3h + lib3h_p2p_protocol) so we should maybe make the time mocking be a separate crate that others could depend on.

crates/lib3h/src/gateway/gateway_transport_send.rs
neonphog: The option was a little awkward, took a compromise approach and made a helper function to at least increase readability.

crates/p2p_protocol/src/p2p.rs
neonphog: :+1:

crates/p2p_protocol/src/p2p.rs
neonphog: :+1:

PR #437, opened 2019-10-01
Make macro to verify successful bind, use in all transport tests.

struktured:
## PR summary

Generalizes the work of #430 so its reusable by other parts of the test suite. Contributes to #414 .

- [x] make options container for ghost test harness that eases configuration of timeout, max iters, delay interval, etc. See #440.
- [x] improve uri api with a builder pattern and some utility functions so we can build a new uri with modified version of an old one easily.
- [x] make macro to try binding multiple times with different ports if necessary (eg `wait_for_bind_result!`)
- [ ] replace existing bind tests with the macro
  - [x] `test_websocket_transport_send_msg`
  - [ ] `test_websocket_transport_reconnect`

## Review checklist 

### DIFF
--- a/crates/lib3h/src/transport/websocket/actor.rs
+++ b/crates/lib3h/src/transport/websocket/actor.rs
@@ -410,10 +410,10 @@ impl
 mod tests {
 
     use super::*;
-    use crate::{tests::enable_logging_for_test, transport::websocket::tls::TlsConfig};
-    use lib3h_ghost_actor::{wait1_for_repeatable_callback, wait_for_message};
+    use crate::{tests::enable_logging_for_test, wait_for_bind_result, transport::websocket::tls::TlsConfig};
     use std::net::TcpListener;
     use url::Url;
+    use lib3h_ghost_actor::wait_for_message;
 
     fn port_is_available(port: u16) -> bool {
         match TcpListener::bind(("127.0.0.1", port)) {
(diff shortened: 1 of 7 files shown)


crates/zombie_actor/src/ghost_test_harness.rs
```diff
@@ -7,29 +7,70 @@
...
+        $crate::wait_did_work!($ghost_actor, options);
     }};
-    ($ghost_actor:ident) => {
-        $crate::wait_did_work!($ghost_actor, true)
-    };
     ($ghost_actor: ident,
-     $should_abort: expr,
-     $timeout : expr
-      ) => {{
+     $options: expr
```
neonphog: woot! so awesome :+1:

crates/lib3h/src/transport/websocket/actor.rs
```diff
@@ -457,76 +459,28 @@ mod tests {
...
+        let init_transport1_address: Lib3hUri =
+            lib3h_protocol::uri::Builder::with_raw_url("wss://127.0.0.1/")
+                .unwrap()
+                .with_port(1024)
+                .build();
+
+        let (_is_match, expected_transport1_address) =
+            wait_for_bind_result!(transport1, t1_endpoint, init_transport1_address);
 
         let port2 = get_available_port(expected_transport1_address.port().unwrap_or_else(|| 0))
```
neonphog: do we want this `unwrap_or_else(|| 0)` here? or should it just fail if it can't find one?

neonphog: (alternately, should it default to 1025 or something instead of 0?)

crates/lib3h_protocol/src/uri.rs
```diff
@@ -127,6 +148,59 @@ impl Lib3hUri {
     }
 }
 
+/// Eases building of a `Lib3hUri` with a fluent api. Users need not
+/// eve: mutate a `Lib3hUri` directly except for efficiency purposes. Instead,
```
zippy:
```suggestion
/// ever mutate a `Lib3hUri` directly except for efficiency purposes. Instead,
```

crates/lib3h/src/transport/websocket/actor.rs
struktured: @neonphog I thought zero was good choice so it dynamically picks something but open to any behavior you think makes sense.

struktured:
> options = excellent.
> 
> Question, should we deprecate MAX_ITERS in favor of a timeout always?

I thought about this and personally I like the extra degree of control. It is actually used to optimize some nested calls of `wait_did_work!`  as well. We can just crank this default value up really high if there is a concern it is breaking things.

struktured:
> options = excellent.
> 
> Question, should we deprecate MAX_ITTERS in favor of a timeout always?

I thought about this and it has a couple advantages to leave it. 
 - avoids having to ever write `loop {`, thus easier to debug logical errors in practice
 - some macros  call `wait_did_work` internally but there is no need to timeout because the parent for loops were already managing a the timeout (eg `wait_for_message`).
 - it can verify invariants in the processing engine- eg. it shouldn't require more than `N` calls for some particular test to pass (when this determinism is possible).

PR #447, opened 2019-10-03
Flakiness fix

ddd-mtl:
longer gossip and timeout interval + more wait_engine_wrapper_until_no_work()

### DIFF
--- a/crates/lib3h/src/engine/space_layer.rs
+++ b/crates/lib3h/src/engine/space_layer.rs
@@ -66,7 +66,6 @@ impl<'engine> GhostEngine<'engine> {
         for (chain_id, mut space_gateway) in space_gateway_map.drain() {
             space_gateway.process(self)?;
             let request_list = space_gateway.drain_messages();
-            debug!("drained: {:?}", request_list);
             did_work = did_work || request_list.len() > 0;
             space_outbox_map.insert(chain_id.clone(), request_list);
             self.space_gateway_map.insert(chain_id, space_gateway);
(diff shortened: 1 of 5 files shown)


crates/lib3h/src/engine/space_layer.rs
```diff
@@ -66,15 +66,15 @@ impl<'engine> GhostEngine<'engine> {
...
+            // debug!("drained: {:?}", request_list);
             did_work = did_work || request_list.len() > 0;
             space_outbox_map.insert(chain_id.clone(), request_list);
             self.space_gateway_map.insert(chain_id, space_gateway);
         }
         // Process all space gateway requests
         for (chain_id, request_list) in space_outbox_map {
             for request in request_list {
-                debug!("process {:?} {:?}", chain_id, request);
+                // debug!("process {:?} {:?}", chain_id, request);
```
neonphog: do we want to switch these to `trace!` instead of commenting them out?

crates/lib3h/tests/integration_test.rs
```diff
@@ -111,8 +111,8 @@ fn setup_memory_node(name: &str, agent_id_arg: Address, fn_name: &str) -> NodeMo
         work_dir: PathBuf::new(),
         log_level: 'd',
         bind_url: Lib3hUri::with_memory(format!("{}/{}", fn_name, name).as_str()),
-        dht_gossip_interval: 1500,
-        dht_timeout_threshold: 5000,
+        dht_gossip_interval: 5000,
```
zippy: I don't think the timeout is the issue.  On my machine I changed it back to:
        dht_gossip_interval: 500,
        dht_timeout_threshold: 3005,

and the test still stops being flaky.  I'm guessing that what really fixed it was the extra set of `wait_engine_wrapper_until_no_work`

I would suggest not raising these unless we really have to.

Issue #288, opened 2019-09-05
Ghost Actor Periodic Task


zippy:
Currently, there are a couple places where we are manually checking things on a timeout... It'd be nice to register a periodic-type task with a tracker somewhere that could automagically get invoked appropriately. See TransportEncoding - the pending_received_data and pending_send_data buffers fill up, and may never empty if handshake fails.

PR #452, opened 2019-10-03
manual ghost actor code - prep for code generation

neonphog:
## Review checklist 


crates/ghost_actor/src/ghost_system.rs
```diff
@@ -0,0 +1,284 @@
...
+            process_recv,
+            process_queue: Vec::new(),
+        }
+    }
+
+    /// first, check for new process functions,
+    /// then, actually loop through the queued process functions
+    /// keep them for next time if they return true
+    fn process(&mut self) -> GhostResult<()> {
+        while let Ok(item) = self.process_recv.try_recv() {
```
struktured: Is there a chance this won't terminate? If so we should cut this off at some finite limit.

struktured: It also suffers from making the queue arbitrarily large.

neonphog: mmm... As far as the system is concerned, any process callback waiting in this queue *must* be processed, or the system won't function properly.

As far as code errors, like infinite loops that create ghost actors, etc... I was thinking it might be interesting at some point to have some configurable "warning" level metrics, like...

- if there are > 1,000,000 actors in the system, or
- the last process() was called > 100ms ago, etc

crates/ghost_actor/src/ghost_protocol.rs
```diff
@@ -0,0 +1,169 @@
...
+            true
+        } else {
+            false
+        }
+    }
+}
+
+/// when a protocol handler received a callback for fulfilling a request
+/// it will follow this signature
+pub type GhostHandlerCb<'lt, T> = Box<dyn FnOnce(T) -> GhostResult<()> + 'lt + Send + Sync>;
```
ddd-mtl: I'd renamed to `GhostHandlerCallback` as to not confuse with `GhostHandler`

crates/ghost_actor/src/ghost_system.rs
```diff
@@ -0,0 +1,284 @@
...
+    }
+
+    pub fn next_run_delay_ms(&self) -> u64 {
+        self.next_run_delay_ms
+    }
+
+    pub fn set_next_run_delay_ms(mut self, next_run_delay_ms: u64) -> Self {
+        self.next_run_delay_ms = next_run_delay_ms;
+        self
+    }
```
ddd-mtl: Do you really need mutating "instructions" ?

crates/ghost_actor/src/ghost_tracker.rs
```diff
@@ -0,0 +1,393 @@
...
+        Self { timeout_ms: None }
+    }
+}
+
+impl GhostTrackerBookmarkOptions {
+    pub fn set_timeout_ms(mut self, timeout_ms: u64) -> Self {
+        self.timeout_ms = Some(timeout_ms);
+        self
+    }
+}
```
ddd-mtl: Same here. Do we really need mutating Options?

crates/ghost_actor/src/ghost_tracker.rs
```diff
@@ -0,0 +1,393 @@
...
+            .expect("can enqueue processor");
+
+        Self {
+            _inner: inner,
+            send_inner,
+        }
+    }
+
+    /// register a periodic task
+    pub fn periodic_task(
```
ddd-mtl: Suggest renaming to `add_periodic_task()`

crates/ghost_actor/src/ghost_tracker.rs
```diff
@@ -0,0 +1,393 @@
...
+    pub fn bookmark(
+        &mut self,
+        span: Span,
+        cb: GhostResponseCb<'lt, X, T>,
+    ) -> GhostResult<RequestId> {
+        self.bookmark_options(span, cb, GhostTrackerBookmarkOptions::default())
+    }
+
+    /// register a callback, using a specific timeout instead of the default
+    pub fn bookmark_options(
```
ddd-mtl: Suggest renaming to `bookmark_with_options()`

crates/ghost_actor/src/ghost_protocol.rs
neonphog: rust seems to like its shorthand... `fn` `impl` etc... I'd be cool changing this, but there's a bunch of other `...Cb` types in here, we should probably pick a pattern and stick with it.

crates/ghost_actor/src/ghost_system.rs
neonphog: This just allows a builder pattern, so when we're using it, we can do things like:

```rust
return Ok(GhostProcessInstructions::default().set_next_run_delay_ms(50))
```

crates/ghost_actor/src/ghost_tracker.rs
neonphog: similar

```rust
return Ok(GhostTrackerBookmarkOptions::default().set_timeout_ms(0))
```

crates/ghost_actor/src/ghost_tracker.rs
neonphog: ended up pulling out the periodic thing for now... updated https://github.com/holochain/lib3h/issues/288 with a hint about this name

PR #454, opened 2019-10-04
SpaceHash newtype

ddd-mtl:
As specified [here](https://hackmd.io/CcOBj5uEQxeRr_Z8Ciw5Ow?view).

### DIFF
--- a/crates/lib3h_protocol/src/types.rs
+++ b/crates/lib3h_protocol/src/types.rs
@@ -1,5 +1,5 @@
 use holochain_persistence_api::hash::HashString;
-use std::{convert::TryInto, fmt};
+use std::fmt;
 
 //--------------------------------------------------------------------------------------------------
 // SpaceHash: newtype for HashString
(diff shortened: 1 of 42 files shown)


crates/lib3h_protocol/src/types.rs
```diff
@@ -0,0 +1,45 @@
+use holochain_persistence_api::hash::HashString;
+use std::fmt;
+
+//--------------------------------------------------------------------------------------------------
+// SpaceHash: newtype for HashString
+//--------------------------------------------------------------------------------------------------
+
+#[derive(PartialOrd, PartialEq, Eq, Ord, Clone, Debug, Serialize, Deserialize, Default, Hash)]
+pub struct SpaceHash(HashString);
+
```
struktured: Consider also using [shrinkwraprs](https://lib.rs/crates/shrinkwraprs). This would be useful if we need to treat them as `HashString`s quite often and use whatever api `HashString` provides.

PR #456, opened 2019-10-07
EntryHash and AspectHash

ddd-mtl:

### DIFF
--- a/crates/lib3h/src/dht/mod.rs
+++ b/crates/lib3h/src/dht/mod.rs
@@ -16,23 +16,23 @@ pub mod tests {
     use lib3h_ghost_actor::prelude::*;
     use lib3h_protocol::{
         data_types::{EntryAspectData, EntryData},
+        types::*,
         uri::Lib3hUri,
-        Address,
     };
 
     lazy_static! {
         /// CONSTS
         /// Entries
-        pub static ref ENTRY_ADDRESS_1: Address = "entry_addr_1".into();
-        pub static ref ENTRY_ADDRESS_2: Address = "entry_addr_2".into();
-        pub static ref ENTRY_ADDRESS_3: Address = "entry_addr_3".into();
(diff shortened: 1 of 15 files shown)


crates/lib3h_protocol/src/types.rs
```diff
@@ -43,3 +43,89 @@ impl SpaceHash {
         SpaceHash(HashString::new())
     }
 }
+
+//--------------------------------------------------------------------------------------------------
+// EntryHash: newtype for HashString
+//--------------------------------------------------------------------------------------------------
+
+#[derive(PartialOrd, PartialEq, Eq, Ord, Clone, Debug, Serialize, Deserialize, Default, Hash)]
+pub struct EntryHash(HashString);
```
neonphog: Can you also do Shrinkwrap and shrinkwrap(mutable) a'la https://github.com/holochain/lib3h/blob/develop/crates/zombie_actor/src/lib.rs#L22-L23 so we get AsRef and friends?

crates/lib3h/src/dht/mod.rs
```diff
@@ -86,8 +86,8 @@ pub mod tests {
 
     #[allow(non_snake_case)]
     fn create_EntryData(
-        entry_address: &Address,
-        aspect_address: &Address,
+        entry_address: &EntryHash,
+        aspect_address: &AspectHash,
```
neonphog: do we want to rename the fields to entry_hash && aspect_hash? I'm not sure I have an opinion either way, just asking the question.

ddd-mtl: no we dont. No point in specifying type in variable name. Its used as an address so name is fine

crates/lib3h_protocol/src/types.rs
ddd-mtl: shrinkwrap(mutable) doesnt work as HashString would need to derive it also I believe.
Don't see the need anyway as there is no need to mutate hashes, no?

PR #444, opened 2019-10-02
adds test_mirror_suite for testing mirroring on multiple nodes

zippy:
## PR summary

Creates a test suite for spinning up n of nodes, and then sees if when one node authors an entry, that all other nodes get the entry gossiped

## Review checklist 
- [ ] The story has unit or integration tests
- [ ] No new bugs, and any tech-debt is identified and justified
- [ ] There is enough API documentation (how to use)
- [ ] There is enough code documentation (how the code works)

PR #461, opened 2019-10-08
networkHash

ddd-mtl:

### DIFF
--- a/crates/lib3h/src/message_encoding.rs
+++ b/crates/lib3h/src/message_encoding.rs
@@ -3,7 +3,7 @@
 use crate::error::{Lib3hError, Lib3hResult};
 use detach::prelude::*;
 use lib3h_ghost_actor::prelude::*;
-use lib3h_protocol::data_types::Opaque;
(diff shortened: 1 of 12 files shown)


struktured:
This would help resolve a conflict in #444 so let's get this in!

crates/lib3h/src/message_encoding.rs
```diff
@@ -42,15 +42,27 @@ pub mod encoding_protocol {
...
     pub enum RequestToChild {
-        Decode { payload: Opaque },
-        EncodeHandshake { space_address: String, id: String },
-        EncodePayload { payload: Opaque },
+        Decode {
+            payload: Opaque,
+        },
+        EncodeHandshake {
+            network_or_space_address: Address,
+            id: String,
```
neonphog: I'm totally cool with this as a merge-able first cut. But these are either network_hash + node_pubkey or space_hash + agent_pubkey. I'm not sure the best way to handle that for clarity.

ddd-mtl: yeah, not ideal, but a step better than having:
`space_address: self.config.network_id.id.clone()`

PR #467, opened 2019-10-09
add disconnect transport error kind  and implemnt for memory and wss tra…

zippy:
#  PR summary

*NOTE: * DO NOT MERGE DEVELOP INTO THIS BRANCH.  CURRENTLY CORE AND SIM2H RELY ON IT WITHOUT THE TYPE CHANGES THAT ARE ON THE TIP OF DEVELOP

Currently disconnection events were not consistently being bubbled up to the parent of a transport.  In our sim2h we need to know where this happens to clean up connections.  We found places where rude disconnects ( e.g. Protocol(Connection reset without closing handshake) ) were occurring and were handled as errors, but regular wss connection closed events weren't being returned.

Promoted Disconnect and Unbind to their own events in the transport protocol.

This PR includes a test for in-memory, but not yet for wss.

## Review checklist 


neonphog:
I currently see two possibilities for carrying this forward.

### 1 - We continue with this multi-level thread-pool concept

If higher levels need to know about a disconnect, I think it makes more sense to bubble it up as a disconnect event (which should contain the Url) similar to how the IncomingConnection event is used.

Eventually we're going to have to put in a modicum of pooling logic at all the levels.

### 2 - We diverge our modules into two separate entities

Something like RawTransport (memory, wss, etc) and ConnectionPool (which Gateway would implement). We would need to add a new wrapper struct (or potentially integrate this into the multiplexer route handling?) that could let the higher level space gateway take something that was in-fact the transport gateway, and make it look like a RawTransport.

The RawTransport could then have more direct commands for terminating / creating / listing connections, while the ConnectionPool / Gateway would manage handshaking / creating / terminating connections.

crates/lib3h/src/transport/protocol.rs
```diff
@@ -29,6 +29,10 @@ pub enum RequestToChildResponse {
 
 #[derive(Debug, Clone)]
 pub enum RequestToParent {
+    // we have become unbound from our interface
+    Unbind(Lib3hUri),
+    // a connection has been dropped
+    Disconnect(Lib3hUri),
     // TODO remove `uri` field once we have refactored how we handle Connection/Disconnection
     ErrorOccured {
```
neonphog: Is ErrorOccured used anymore? Should it be removed?

neonphog: n/m ... saw below

PR #472, opened 2019-10-10
Switch to a wrapper around parking_lot::Mutex for Mutex

neonphog:
## PR summary

No actual functionality changes here, just switching Mutex -> GhostMutex which is a wrapper around parking_lot::Mutex which gives us much better control of timeouts when attempting to lock.

## Review checklist 

### DIFF
--- a/crates/ghost_actor/src/ghost_mutex.rs
+++ b/crates/ghost_actor/src/ghost_mutex.rs
@@ -1,0 +1,197 @@
+extern crate lock_api;
+extern crate parking_lot;
+
+use std::{
+    borrow::{Borrow, BorrowMut},
+    convert::{AsMut, AsRef},
+    ops::{Deref, DerefMut},
+};
+
+#[must_use]
+pub struct GhostMutexGuard<'lt, T: ?Sized + 'lt> {
+    guard: Option<parking_lot::MutexGuard<'lt, T>>,
+}
+
+impl<'lt, T: ?Sized + 'lt> Drop for GhostMutexGuard<'lt, T> {
+    fn drop(&mut self) {
+        match std::mem::replace(&mut self.guard, None) {
+            Some(guard) => {
+                parking_lot::MutexGuard::unlock_fair(guard);
+            }
+            None => (),
+        }
+    }
+}
+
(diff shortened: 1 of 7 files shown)


crates/ghost_actor/src/ghost_mutex.rs
```diff
@@ -0,0 +1,197 @@
...
+    pub fn into_inner(self) -> T {
+        self.mutex.into_inner()
+    }
+}
+
+impl<T: ?Sized> GhostMutex<T> {
+    pub fn lock(&self) -> GhostMutexGuard<'_, T> {
+        match self
+            .mutex
+            .try_lock_for(std::time::Duration::from_millis(10))
```
dymayday: That's a small timeout no ? Could you tell me how you found this to be optimal in your opinion ?

crates/ghost_actor/src/ghost_mutex.rs
```diff
@@ -0,0 +1,197 @@
...
+    use std::sync::Arc;
+
+    #[test]
+    fn it_should_debug() {
+        let a = GhostMutex::new(true);
+        assert_eq!("GhostMutex { inner: true }", &format!("{:?}", a));
+    }
+
+    #[test]
+    fn it_should_be_fair_ish() {
```
dymayday: I'm not sure to understand what this test is about :thinking:

crates/ghost_actor/src/ghost_mutex.rs
neonphog: For the moment, we are running in single threaded mode, so the only time a lock should be a problem is if we lock it twice from the same thread... in which case we want it to panic! so we can fix the circular dependency. But 1 second would be fine from a human time perspective and that might be more appropriate if we ever went multi-threaded too... will update.

crates/ghost_actor/src/ghost_mutex.rs
neonphog: With the rust system Mutex, if a thread unlocks a mutex, then requests a lock again before the next system switch it will receive the lock, leading to a single thread preference.

Parking lot has a fair unlock option that tries to let other threads get a chance to run.

PR #473, opened 2019-10-11
setup ghost endpoint seed/plant/regress

neonphog:
## PR summary

Preparation for being able to pass actors around / use them as child actors.

- adds an EndpointSeed concept that can be `plant`ed : )
- renames EndpointRef which I was never super happy with to EndpointFull ... happy for other suggestions
- EndpointFull instances can be `regress`ed back into seeds if you want to move it to another actor

## Review checklist 


neonphog:
> Can you document the critical functions such as plant and regress?

:+1: - done

PR #482, opened 2019-10-15
Ghost Actors can manage Sub Actors

neonphog:
## PR summary

Removes the awkward Inflator concept in favor of private "DeepRefs" and GhostActorSystem refs that know how to "plant" things with your local context.

## Review checklist 


crates/ghost_actor/src/ghost_actor.rs
```diff
@@ -263,97 +318,15 @@ pub trait GhostEndpoint<'lt, X: 'lt + Send + Sync, P: GhostProtocol> {
 
 /// Describes an actor that can be used within the "Ghost" actor system
 pub trait GhostActor<'lt, P: GhostProtocol, A: GhostActor<'lt, P, A>>: Send + Sync {
+    /// This is an indication that it is ok to begin processing
+    /// before actor_init is called, our self context may not be initialized
+    fn actor_init(&mut self) -> GhostResult<()> {
```
zippy: `actor_init` sounds like a command.  I think this would be better named `is_initialized` or something like that?

ddd-mtl: `can_init_actor()` and change the return type to bool

neonphog: updated comment

PR #486, opened 2019-10-17
rrdht_utils

neonphog:
## PR summary

Provides RRDHT utilities for:
 - calculating a u32 "Location" for a given agent or node id
 - manipulating "Locations" using wrapping (circular) math
 - comparing the "forward" distance between two locations (i.e. always from the first point to the second point even if that means wrapping around the circle)
 - constructing / accessing data about "Arcs", that is, a start location and length and/or a center point and a radius
 - and finally determining whether a "Location" is inside an "Arc"

## Review checklist 

### DIFF
--- a/crates/lib3h/src/rrdht_util/arc.rs
+++ b/crates/lib3h/src/rrdht_util/arc.rs
@@ -4,6 +4,17 @@ pub const ARC_LENGTH_MAX: u64 = 0x100000000;
 pub const ARC_RADIUS_MAX: u32 = 0x80000001;
 
 /// An rrdht "arc" indicates a range on the u32 "location" circle
+/// In general, to implement the rrdht sharding algorithm, new `Arc`s
+/// will need to be created using `Arc::new_radius(center, radius)`.
+/// The `center` parameter for a storage arc, for example, will be the
+/// agent's u32 "Location", and the `radius` (read: half arc-length)
+/// will be how large an arc around that center location the agent
+/// is claiming to store.
+/// - 0 indicates they are storing nothing
+/// - 1 indicates they are storing only those entries that hash to the
+///     exact same u32 location value
+/// - 2 indicates they store that same u32 value + one on each side, etc.
+/// - ARC_RADIUS_MAX indicates they claim to store all values.
 #[derive(Clone, Copy, PartialEq, Eq)]
 pub struct Arc {
     start: Location,
@@ -111,7 +122,7 @@ impl Arc {
 
     /// returns `true` if given location is within this arc
     pub fn contains_location(&self, location: Location) -> bool {
-        self.start.forward_distance_to(location) < self.length
+        u64::from(self.start.forward_distance_to(location)) < self.length
     }
 }
 
--- a/crates/lib3h/src/rrdht_util/arc.rs
+++ b/crates/lib3h/src/rrdht_util/arc.rs
@@ -49,7 +49,10 @@ impl Arc {
         assert_eq!("32r", pre);
         let loc = u32::from_str_radix(&repr[3..11], 16).expect("can parse hex");
         let len = u64::from_str_radix(&repr[12..], 16).expect("can parse hex");
-        Self { start: loc.into(), length: len }
+        Self {
+            start: loc.into(),
+            length: len,
+        }
     }
 
     /// the start position for this arc
--- a/crates/lib3h/src/rrdht_util/arc.rs
+++ b/crates/lib3h/src/rrdht_util/arc.rs
@@ -49,6 +49,32 @@ impl Arc {
         unimplemented!();
     }
 
+    /// the start position for this arc
+    pub fn start(&self) -> Location {
+        self.start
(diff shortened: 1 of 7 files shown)


crates/lib3h/src/rrdht_util/arc.rs
```diff
@@ -0,0 +1,308 @@
...
+}
+
+impl std::fmt::Debug for Arc {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_tuple("Arc").field(&format!("{}", self)).finish()
+    }
+}
+
+impl Arc {
+    /// construct a new Arc instance given a start Location and a length
```
struktured: Add note about length possibly being clipped to a max value.

crates/lib3h/src/rrdht_util/arc.rs
```diff
@@ -0,0 +1,308 @@
...
+        let start = center - Location::from(radius - 1);
+        let length = u64::from(radius) * 2 - 1;
+        Arc::new(start, length)
+    }
+
+    /// construct a new Arc from canonical BITSrHEX:HEX format
+    pub fn new_repr(repr: &str) -> Self {
+        let pre = &repr[0..3];
+        assert_eq!("32r", pre);
+        let loc = u32::from_str_radix(&repr[3..11], 16).expect("can parse hex");
```
struktured: This error seems easy to occur - I would suggest encoding the input string at the very least in the `expect` and consider changing it to an `unwrap_or_else` clause Same with `len` as below this line.

crates/lib3h/src/rrdht_util/arc.rs
```diff
@@ -0,0 +1,308 @@
...
+        }
+        self.start + (self.radius() - 1).into()
+    }
+
+    /// the radius of this arc
+    pub fn radius(&self) -> u32 {
+        if self.length == 0 {
+            return 0;
+        }
+        (self.length / 2 + 1) as u32
```
struktured: Is the +1 to account for the remainder when lengths are odd? If so this seems wrong for even lengths. But maybe this is an offset instead. Regardless a comment might be helpful here.

neonphog: This is actually to reverse the `- 1` when [creating an arc](https://github.com/holochain/lib3h/blob/rrdht-util/crates/lib3h/src/rrdht_util/arc.rs#L42) based off a center point and a "radius" (erm... half the length?). We this value to mean not storing anything when zero and only store the exact center point when 1, which requires this subtraction / addition.

Will try to write a less ramble-y comment : )

crates/lib3h/src/rrdht_util/arc.rs
```diff
@@ -0,0 +1,336 @@
...
+            return Arc::new(center, 0);
+        }
+
+        let radius = std::cmp::min(radius, ARC_RADIUS_MAX);
+
+        // we need to subtract 1 from the radius in the following
+        // calculations to ensure that:
+        // - 0 means no coverage
+        // - 1 means coverage only including the center point
+        // - any thing obove expand out to the sides
```
lucksus: This comment looks like it could make sense to promote it to a rustdoc comment on the whole type..

crates/lib3h/src/rrdht_util/location.rs
```diff
@@ -0,0 +1,119 @@
...
+/// i.e. are mapped onto a circle
+#[derive(Shrinkwrap, Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[shrinkwrap(mutable)]
+pub struct Location(pub Wrapping<u32>);
+
+impl Location {
+    /// get the distance from this location to an `other` location
+    /// this distance is processed forward taking wrapping into account
+    #[allow(clippy::trivially_copy_pass_by_ref)]
+    pub fn forward_distance_to(&self, other: Location) -> u64 {
```
lucksus: I wonder, do we not need the symmetric distance, i.e. distance taking both directions around the circle into account?

crates/lib3h/src/rrdht_util/arc.rs
neonphog: :+1: - done

crates/lib3h/src/rrdht_util/location.rs
neonphog: I don't have a current mathematical need for that, but it should be easy to add if we come across one : )

neonphog:
@lucksus 
> should this code live in its own rrDHT crate?

It certainly could. I don't have a strong opinion. There's some overhead in terms of crating a standalone error type...

crates/lib3h/src/rrdht_util/arc.rs
```diff
@@ -0,0 +1,347 @@
...
```
zippy:
```suggestion
        // - any thing above expand out to the sides
```

PR #487, opened 2019-10-18
rrdht r-value / recommended storage arc radius utility

neonphog:
## PR summary

Provides two utility functions for working with "R" (redundancy) Values and storage arc radii.

- `interpolate_r_value_for_given_arc` allows you to estimate the total r-value with a subset of sample data (in a sharded context this might be the data you are actually storing).
- `get_recommended_storage_arc_radius` as an agent, gives you a recommendation for what your storage arc should be based on our estimated knowledge of network conditions.

For SIM2H - we may have to fudge some of the data until we have it (like uptime)

## Review checklist 


jmday:
This just makes me really happy.

That is all.

crates/lib3h/src/rrdht_util/r_value.rs
```diff
@@ -0,0 +1,337 @@
...
+        // update our running percentage covered
+        running_pct_total += relative_pct_of_space_covered;
+    }
+
+    let pct_of_space_covered: f64 =
+        peer_record_set.arc_of_included_peer_records.length() as f64 / ARC_LENGTH_MAX as f64;
+
+    // we are only sampling the pct_of_space_covered
+    // we need to grow our sample (interpolate) to the rest of the space
+    running_pct_total * (1.0 / pct_of_space_covered)
```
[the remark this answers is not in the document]
neonphog: Let's say we've got two buckets of equal size (half the space), bucket A is known, and bucket B is unknown.

Bucket A has a single item in it that covers 1/3 of the TOTAL space.

If we interpolate such that we assume bucket B has an identical item that also covers 1/3 of the total space, we should assume that 2/3 of the total space is covered, and this math checks out:

```
(1 / 3) * (1 / 0.5) == (2 / 3)
```

Thus, we do want to know the total percentage coverage for the items within our small sample, and then multiply for assumed counterparts in the unknown portion of the circle.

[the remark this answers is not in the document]
jmday: Considering the storage arc not as "storage" but "the portion of the arc for which I know agents," the current logic makes sense to me. Basically, @neonphog is calculating just how much "expected availability" of data across the entire DHT is provided by the agents within that arc. Interpolation by scaling up WRT the portion of the DHT the arc covers should provide a valid estimate of the "expected availability" of data in the entire DHT.

I sense a clipping approach could provide useful insight, too, but would require much more complex math to interpolate a valid DHT-wide result.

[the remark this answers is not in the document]
jmday: Also, I'd suggest replacing the term `pct` with something akin to `DHT_coverage` ... aggregate expected coverage beyond 100% is quite likely here (consider an agent with a storage arc > .5 * DHT_size), which feels funny to conceptualize as a "percent"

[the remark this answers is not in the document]
neonphog: Hehe, indeed, if you look at my unit-tests, they are looking for an "R" value between 25 and 50... aka 2500% to 5000% : )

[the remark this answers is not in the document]
lucksus: The problem I see: there is no guarantee that all `peer_records` are completely included in the `arc_of_included_peer_records`. If one is not, the calculation is off -> solution: clipping `record.storage_arc` with `arc_of_included_peer_records` (i.e. just calculate the intersection of those arcs).

The alternative would be to regard this constraint as part of the contract of this function. There is no problem if it is never used on data that does not adhere to this implicit contract. I believe that the cost of adding above mentioned clipping is far less then potential debugging in the future.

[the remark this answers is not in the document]
jmday: My understanding of the intended rrDHT sharding approach is that it requires an agent to maintain awareness of ALL active peers within its storage arc. While there may be short periods where it doesn't know all peers, the impact on the calculation of expected coverage should be negligible... but @neonphog may have updated this intention since I last had a deep dive?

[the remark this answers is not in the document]
neonphog: @lucksus - Let me try to re-phrase your comment - please tell me if I'm not hearing you correctly.

Let's say we have an arc with only 10 unit positions within it. That is, your "Location" can be 0-9 inclusive. Given the same equal sized A and B buckets above, let's define A as 0-4 inclusive, and B as 5-9. That is they, both have 5 unit length.

Now, our `arc_of_included_peer_records` is A (0-4). The peer we know about is located at location 0, with a radius of 2, so, the start of its arc is actually 9, and the inclusive end of its arc is 1. (its length is 4 exclusive).

That 9 position is outside our `arc_of_included_peer_records` so you are suggesting we clip the arc to match, which would cause us to treat it as if it were covering 0 to 1.

---

I submit that we don't actually care about the specific positioning of these agent's arcs. We are dealing in aggregate here, and while some of the arcs we know about may extend beyond our range, it is also true that some of the arcs we don't know about (because their center location is beyond our storage arc) actually extend into our range as well.

If we were to clip our entries to our arc, it would actually make our calculations less accurate because we wouldn't be taking this second case into account.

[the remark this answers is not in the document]
neonphog: Another way to get the same result might be to:
- step 1 - figure out the average total coverage for all known nodes.
- step 2 - guess how many total nodes there are based on our known node count.
- step 3 - guess the total coverage by multiplying steps 1 and 2.

This method, I think, makes it extra clear that we wouldn't want to clip, because then our average coverage would be incorrect. I *think* what's currently implemented gives the same result, but perhaps it would be better to switch to this for code clarity.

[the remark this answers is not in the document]
neonphog: @jmday / @lucksus - removed pct references and implemented alternate calculation strategy. Please provide feedback.

crates/lib3h/src/rrdht_util/r_value.rs
```diff
@@ -0,0 +1,335 @@
...
+    pub fn arc_of_included_peer_records(mut self, arc_of_included_peer_records: Arc) -> Self {
+        self.arc_of_included_peer_records = arc_of_included_peer_records;
+        self
+    }
+
+    /// push a new PeerRecord into this record set
+    pub fn push_peer_record(mut self, peer_record: RValuePeerRecord) -> Self {
+        if !self
+            .arc_of_included_peer_records
+            .contains_location(peer_record.storage_arc.center())
```
jmday: Suggest using something like `location` instead of `center` in case we need to shift to arcs that are not symmetric around a peer address location. For now it can all be the mid-point, but I sense the future is likely to reveal there are reasons to have asymmetric storage arcs.

crates/lib3h/src/rrdht_util/r_value.rs
```diff
@@ -0,0 +1,335 @@
...
+        {
+            panic!("peer record does not fit within arc_of_included_peer_records");
```
jmday: Suggest "peer record location is not located within arc_of_included_peer_records" instead. A "does not fit" error could be misinterpreted as the arc of of the peer record not being contained by the arc_of_included_peer_records. I sense this could lead to some needless & misled debugging efforts.

crates/lib3h/src/rrdht_util/r_value.rs
```diff
@@ -0,0 +1,335 @@
...
+        }
+        self.peer_records.push(peer_record);
+        self
+    }
+}
+
+/// Interpolate qualified redundancy in the network given our current
+/// knowledge of peers' claimed storage arcs.
+/// For a given storage arc (should match *this* agent's storage arc),
+/// if all the peers we know about are actually storing what they claim to be
```
jmday: Probably important to state clearly that "we" (i.e. the focal agent) are obligated to know about all peers within our storage arc. This is an important assumption for the statistical inference we're making for valid interpolation. Probably not necessary to state the eventually consistent nature of the peer list containing all peers within the storage arc.

crates/lib3h/src/rrdht_util/r_value.rs
```diff
@@ -0,0 +1,335 @@
...
+pub fn interpolate_r_value_for_given_arc(peer_record_set: &RValuePeerRecordSet) -> f64 {
+    let mut running_avg_coverage = 0.0_f64;
+
+    for (i, record) in peer_record_set.peer_records.iter().enumerate() {
+        // what is the total coverage factor for this record?
+        let coverage =
+            (record.storage_arc.length() as f64 / ARC_LENGTH_MAX as f64) * record.uptime_0_to_1;
+
+        // mix it into our running average
+        running_avg_coverage = (running_avg_coverage * i as f64 + coverage) / (i as f64 + 1.0);
```
jmday: I have a suggestion for a different approach to this calculation (based on previous experience trying this in a similar context)...
Welford Stat Tuples (WST):
  - Provides for accumulation of count, expected value, and a variance measure. And it may be really (um, eventually REALLY) useful to capture the variance in addition to the expected coverage.
  - WSTs allow for almost-lossless O(1) updates with additions/deletions. The expected value method currently used will experience significant loss of precision because it will accumulate floating point errors. (you'd think they'd average out, but they don't... and the * i and / i just make it worse unfortunately)
  - WSTs are associative & commutative. Thus (in addition to single additions/deletions) partial statistical results can be aggregated (or dis-aggregated) without loss. This means, we could optimize this calculation when/where DHT shards overlap (e.g. based on a binary-based hash location hierarchies) by sharing partial results... This is a future optimization I want to discuss with you @neonphog, but WSTs will prepare us well.

crates/lib3h/src/rrdht_util/r_value.rs
```diff
@@ -0,0 +1,335 @@
...
+    }
+
+    // given our known arc, estimate the total node count
+    let est_node_count = peer_record_set.peer_records.len() as f64
+        * (1.0
+            / (peer_record_set.arc_of_included_peer_records.length() as f64
+                / ARC_LENGTH_MAX as f64));
```
jmday: Couldn't this just be
`* (ARC_LENGTH_MAX as f64 / peer_record_set.arc_of_included_peer_records.length() as f64`
?

To me, it's just clearer to use the inverse ratio total DHT arc vs. our storage arc to scale up. I found myself doing some mental gymnastics to realize what was being achieved here with double division.

crates/lib3h/src/rrdht_util/r_value.rs
```diff
@@ -0,0 +1,335 @@
...
+/// on the one hand, if we have enough storage and network bandwidth,
+/// we might want to set it as high as possible. On the other hand, we
+/// might want to conserve resources / power. If our network is healthy,
+/// we can scale back a bit to conserve, if it is in danger, we may want
+/// to scale up for safety. If it is anywhere in the middle, we probably
+/// want to maintain stability so we're not thrashing our resources.
+///
+/// Given current network conditions (within the sample slice given)
+/// this function will return a recommended storage arc radius.
+/// RADIUS not length. Please use `Arc::new_radius(0.into(), radius).length()`
```
jmday: Using an arc radius (which could be a rather confusing concept in itself since it's not referring to the radius of the circle which the arc is on, but rather the 1/2 size of an arc that is symmetric around a center location) limits us to using symmetric storage arcs.

I want us to eventually be free from the assumption that a storage arc is symmetric around the address location of the peer. I sense it's a special case that we're committing to and there will be benefits if we can more easily explore the general case problem later. Are there some minor adjustments that could be made here that allow us to more easily transition later?

crates/lib3h/src/rrdht_util/r_value.rs
```diff
@@ -0,0 +1,335 @@
...
+/// if you need to determine the length for a given radius..
+///
+/// If you provide a current_arc_radius, and the network is anywhere near
+/// stable, this function will probably return your current radius back.
+/// Otherwise it will pick a maintenance target right in the middle.
+///
+/// If the network is over replicated, this function may pick a small radius.
+/// If the network is unhealthy, or immature this function may pick a large
```
jmday: Suggest "under replicated" instead of "unhealthy" here. The phrase "network is unhealthy" could be misinterpreted many ways (e.g. too many network partitions, malicious actors, high latency).

crates/lib3h/src/rrdht_util/r_value.rs
```diff
@@ -0,0 +1,335 @@
...
+/// arc radius, up to 100% in the case of a new / immature network.
+///
+/// Also, this is a bit of a naive first implementation, expect the specific
+/// algorithm / heuristics to be updated on an on-going basis.
+pub fn get_recommended_storage_arc_radius(
```
jmday: Suggest `get_recommended_storage_arc_size` and just let the current implementation assume symmetric arcs.

crates/lib3h/src/rrdht_util/r_value.rs
```diff
@@ -0,0 +1,335 @@
...
+    peer_record_set: &RValuePeerRecordSet,
+    target_minimum_r_value: f64,
+    target_maximum_r_value: f64,
+    current_arc_radius: Option<u32>,
```
jmday: Suggest `current_arc_size`

crates/lib3h/src/rrdht_util/r_value.rs
```diff
@@ -0,0 +1,335 @@
...
+) -> u32 {
+    let pct_of_space_covered: f64 =
+        peer_record_set.arc_of_included_peer_records.length() as f64 / ARC_LENGTH_MAX as f64;
+    let interp_total_node_count =
+        peer_record_set.peer_records.len() as f64 * (1.0 / pct_of_space_covered);
```
jmday: These look like good candidates for methods that could be implemented on RValuePeerRecordSet. `pct_of_space_covered` could also be used in the final r_value interpolation calculation within `interpolate_r_value_for_given_arc` above.

crates/lib3h/src/rrdht_util/r_value.rs
```diff
@@ -0,0 +1,335 @@
...
+    if interp_total_node_count < target_minimum_r_value * 2.0 {
```
jmday: Since this is the comparison of a count and something akin to a percent (or a least a desired expectation value), it may be good to add a comment here explaining why this comparison is really what is desired.

crates/lib3h/src/rrdht_util/r_value.rs
```diff
@@ -0,0 +1,335 @@
...
+        // consider this network immature, recommend full coverage
+        return ARC_RADIUS_MAX;
+    }
+
+    let cur_r_value = interpolate_r_value_for_given_arc(peer_record_set);
```
jmday: Suggest `cur_r_value_estimate` instead of `cur_r_value` (which sounds too confident).

crates/lib3h/src/rrdht_util/r_value.rs
```diff
@@ -0,0 +1,335 @@
...
+    let mut count: u64 = 0;
+    let mut rad_avg: f64 = 0.0;
+
+    #[allow(clippy::explicit_counter_loop)]
+    for record in peer_record_set.peer_records.iter() {
+        count += 1;
+
+        let rad = record.storage_arc.radius();
+
+        rad_avg = (rad_avg * (count as f64 - 1.0) + f64::from(rad)) / count as f64;
```
jmday: This looks like the same calculation as in `interpolate_r_value_for_given_arc` but just for storage arc radii. I'd suggest WSTs here too (for all the same reasons), but I sense we could optimize so that this re-calculation doesn't need to occur. Perhaps break out a separate function for calculation from `RValuePeerRecordSet::avg_coverage`? If avg radius is needed for the current symmetric case, it should be valid to 1/2 it, right?

crates/lib3h/src/rrdht_util/r_value.rs
```diff
@@ -0,0 +1,335 @@
...
+    }
+
+    let out_radius: f64 = if cur_r_value < target_minimum_r_value {
+        // our network is unhealthy! let's try to capture more!
```
jmday: Suggest "under replicated" instead of "unhealthy"

crates/lib3h/src/rrdht_util/r_value.rs
```diff
@@ -0,0 +1,335 @@
...
+
+        // let's double the current average for our target mid-point
+        let new_mid_rad = rad_avg * 2.0;
```
jmday: I sense there could be problems with this constant approach. What if we based expansion target on current arc, some factor (e.g. 2x for now, but could be based on relative uptime, available storage, bandwidth, etc. in the future), and the ratio of `cur_r_value / target_minimum_r_value` instead? If no current arc, then probably base off rad_avg (instead of current arc) as a starting point?

crates/lib3h/src/rrdht_util/r_value.rs
```diff
@@ -0,0 +1,335 @@
...
+
+        match current_arc_radius {
+            None => new_mid_rad,
```
jmday: Suggest `rad_avg * (cur_r_value / target_minimum_r_value)` as starting point target... i.e. just target carrying your own weight.

crates/lib3h/src/rrdht_util/r_value.rs
```diff
@@ -0,0 +1,335 @@
...
+            Some(cur) => {
+                let cur = f64::from(cur);
+
+                // try to maintain some stability
+                // we'll accept any current from 1.5 to 3 times the average
+                let new_min_rad = rad_avg * 1.5;
+                let new_max_rad = rad_avg * 3.0;
```
jmday: Curious how this stability approach will play out. I sense there may be value in involving factors related to ratios of cur_r_value with `target_minimum_r_value` and/or `target_maximum_r_value` here and below with over replicated networks.

crates/lib3h/src/rrdht_util/r_value.rs
```diff
@@ -0,0 +1,335 @@
...
+        }
+    } else {
+        // our network is perfect, let's try to keep it this way!
+        match current_arc_radius {
+            None => rad_avg,
+            Some(cur) => f64::from(cur),
+        }
+    };
+
+    // we might be tempted to correct for any shortage in our own uptime here
```
jmday: I sense a correction factor that incorporates uptime could be useful as a direct correlation (rather than inverse). Nodes that are up within the DHT for longer should be storing more... they have more time to accumulate data, they're more invested in the DHT, and if they're returning, they probably already have more DHT data so gossip to update them to current will require less bandwidth. This could be really useful for ensuring new peer entrants don't inundate the DHT with gossip update requests, too.

crates/lib3h/src/rrdht_util/r_value.rs
neonphog: ... since the "location" is not stored in the structure, I'd have no way of determining it if it was not in the center.

That's not to say we can't support that; you can create an arc with start and length, and offset the start however you want... but if we're using the `new_radius` function it seems like the center needs to actually be in the center.

Also, the "radius" concept came about from trying to minimize the data needed to be transferred during gossip. If you know their AgentId - 4 bytes gets you their storage arc, because you know it is centered around the location derived from their AgentId.

crates/lib3h/src/rrdht_util/r_value.rs
neonphog: This does not resolve easily on google ;p - I'm not against it, but it doesn't seem like an MVP thing. This code is just to get us going, we should feel free to update our algorithm whenever there is need.

crates/lib3h/src/rrdht_util/r_value.rs
neonphog: lol : )

crates/lib3h/src/rrdht_util/r_value.rs
neonphog: See my above comment about minimizing gossip data.

I don't feel like I'm baking us into any corners here... we can and will need to alter this api to accept additional metrics for determining the best arc for given agents, so we shouldn't feel like this MVP is the end-all of the conversation.

two other things specifically about non-centered arcs though:
- if we have so few people on our network that the specific positioning of their arc relative to their center location makes any difference, then our network is unhealthy and everyone should hold full arcs anyways
- I think giving agents control of how their arcs are oriented may be too game-able... OH if i just stretch in this direction a bit, I can have a voice in that entries validation!

crates/lib3h/src/rrdht_util/r_value.rs
neonphog: no, the previous avg takes the uptime into consideration, where here I need the literal average arc radius

crates/lib3h/src/rrdht_util/r_value.rs
neonphog: Indeed, there are a billion different ways to do this. Likely most of them are better, but we need somewhere to start.

crates/lib3h/src/rrdht_util/r_value.rs
neonphog: > Nodes that are up within the DHT for longer should be storing more

Only on the time-scale of minutes... hApps that require hours of downloads should hopefully be outliers... I think for the most part, you can assume everyone on the network holds as much data as they plan to.

.. I'm not sure I understand the other bits you mention here... All I was saying is that the "R" value target we calculated already adjusts for uptime - so if we increased the recommendation to account for our own uptime, everybody would slowly be increasing their arcs for no reason.

neonphog:
@jmday - See my comments above - I stopped explicitly commenting about the center-point things and the specific algorithm choices, as this is just intended to be a naive starting point until we can actually start getting some data about what is working and what isn't.

If there's anything you think absolutely needs to be changed before this starting point, totally let me know! : )

crates/lib3h/src/rrdht_util/r_value.rs
jmday: Agreed.

And I totally should've linked something useful (but was on a plane without wifi when responding)!

https://www.johndcook.com/blog/standard_deviation/

crates/lib3h/src/rrdht_util/r_value.rs
jmday: Let's come back to discussing this if we find DHT entry gossip gets to be too intensive. I sense it would be a good way to gracefully serve a large number of new entrants or large number of very infrequent users.

crates/lib3h/src/rrdht_util/r_value.rs
```diff
@@ -0,0 +1,341 @@
+use crate::rrdht_util::{Arc, ARC_LENGTH_MAX, ARC_RADIUS_MAX};
+
+/// builder for our knowledge of a single agent's storage / reachability
+pub struct RValuePeerRecord {
+    storage_arc: Arc,
+    uptime_0_to_1: f64,
```
struktured: What about `uptime_prob: f64` short for _probability_?

struktured: Also ok with `uptime_perc` but then scaling is possibly confusing.

crates/lib3h/src/rrdht_util/r_value.rs
```diff
@@ -0,0 +1,341 @@
...
+/// if all the peers we know about are actually storing what they claim to be
+/// we can interpolate what the whole network "R" value is
+/// also taking into account our experience of those agents' reachability.
+pub fn interpolate_r_value_for_given_arc(peer_record_set: &RValuePeerRecordSet) -> f64 {
+    let mut running_avg_coverage = 0.0_f64;
+
+    for (i, record) in peer_record_set.peer_records.iter().enumerate() {
+        // what is the total coverage factor for this record?
+        let coverage =
+            (record.storage_arc.length() as f64 / ARC_LENGTH_MAX as f64) * record.uptime_0_to_1;
```
struktured: Weighted linearly by uptime. Interesting heuristic but I think this is a great candidate for a ML system that considers uptime percentage as just one of several features (network latency could be another). 

For now I think this is perfectly fine but it definitely needs to be tested, especially around the nearly zero uptime boundary. Is there a paper I haven't read that establishes convergence properties of this network regarding coverage? Intuitively to me how much you store is not so strongly dependent on uptime, but in the above equation it certainly is.

crates/lib3h/src/rrdht_util/r_value.rs
struktured: @jmday we eventually need statistical methodologies weaved into this for initial values and how to produce estimates of R and otherwise. I think the important thing is to ensure the api is flexible enough to incorporate more advanced techniques, especially the adoption of other features to produce estimates.

crates/lib3h/src/rrdht_util/r_value.rs
struktured: Also ok with just leaving this as is.

zippy:
I see this already got merged.  It does look great as MVP.  The one thing I was going to suggest is to pull out some of the stuff in the interpolation loop into for semantic function calls based on a strategy, so that it's clearer from the start how to add strategies later.  But that's not really necessary.  Nice work.

PR #489, opened 2019-10-21
WIP: ghost-actor-v2 - Tracing

ddd-mtl:
Spam of spans

#### Misc
Renamed: `plant_endpoint()` -> `plant_seed()`


crates/ghost_actor/src/ghost_actor.rs
```diff
@@ -224,35 +226,38 @@ impl<
...
             }
-            None => self.send.send((None, message))?,
+            None => self.send.send((span, None, message))?,
         }
         Ok(())
     }
 
     fn priv_process_handle_requests(&mut self, user_data: &mut X) -> GhostResult<()> {
-        while let Ok((maybe_id, message)) = self.recv.try_recv() {
+        while let Ok((span, maybe_id, message)) = self.recv.try_recv() {
```
neonphog: In the manual test_protocol the handlers have Span::fixme()'s in them that would go away if we send this span in through the `trigger` at the end of this function.

crates/ghost_actor/src/ghost_tracker.rs
```diff
@@ -69,14 +70,15 @@ impl<'lt, X: 'lt + Send + Sync, T: 'lt + Send + Sync> GhostTrackerInner<'lt, X,
...
         user_data: &mut X,
         request_id: RequestId,
         data: T,
     ) -> GhostResult<()> {
         match self.pending.remove(&request_id) {
-            None => return Err(GhostErrorKind::RequestIdNotFound("".to_string()).into()),
+            None => return Err(GhostErrorKind::RequestIdNotFound(String::new()).into()),
             Some(entry) => {
-                (entry.cb)(user_data, Ok(data))?;
+                (entry.cb)(span.child("process_handle"), user_data, Ok(data))?;
```
neonphog: so the thought here is that we use the passed in span rather than the one stored in the entry that we use for the timeout?

This is where I'm a little unsure what the best-practices around spans should be. Do we just let that other one drop? do we put a "fulfilled" message on it?

I think this is totally awesome as a first cut, just curious what we should do next here : )

ddd-mtl: Yeah, I'm not so sure adding a span to GhostTrackerEntry is a good idea so I try not to use it unless I have to.

crates/ghost_actor/src/ghost_actor.rs
ddd-mtl: When we talked, I heard you say no need for span in trigger(), so I tried not to, but I can change that no problem.

PR #491, opened 2019-10-21
allow error cloning for ghost actor 2.0

neonphog:
## PR summary

GhostActor 2.0 depends on Protocols being clone-able. This makes Lib3h Error types clone-able.

## Review checklist 

### DIFF
--- a/crates/lib3h/src/error.rs
+++ b/crates/lib3h/src/error.rs
@@ -12,7 +12,7 @@ use std::{error::Error as StdError, fmt, io, result};
 pub type Lib3hResult<T> = result::Result<T, Lib3hError>;
 
 /// An error that can occur when interacting with the algorithm.
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct Lib3hError(Box<ErrorKind>);
 
 impl Lib3hError {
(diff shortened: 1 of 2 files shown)


crates/lib3h/src/error.rs
```diff
@@ -73,6 +73,23 @@ pub enum ErrorKind {
...
 
+impl Clone for ErrorKind {
+    fn clone(&self) -> Self {
+        match self {
+            ErrorKind::GhostError(e) => ErrorKind::GhostError(e.clone()),
+            ErrorKind::Io(e) => ErrorKind::Io(e.kind().into()),
+            ErrorKind::TransportError(e) => ErrorKind::TransportError(e.clone()),
+            ErrorKind::Lib3hProtocolError(e) => ErrorKind::Lib3hProtocolError(e.clone()),
+            ErrorKind::HcId(e) => ErrorKind::HcId(e.clone()),
+            ErrorKind::RmpSerdeDecodeError(e) => ErrorKind::Other(format!("{:?}", e)),
```
neonphog: This is really the only questionable part... but the plan is to move away from RmpSerde so it shouldn't be a problem in the future, and in the mean time we still at least get the error data.

